PR #228, opened 2022-05-23
Add type checker to Pulumi YAML

iwahbe:
Adds a type checker for intermediate values. The type checker currently errors out programs that are invalid (do not pass the type check).

We might want to consider providing warnings instead of errors until we can test the type checker against more input.


pkg/pulumiyaml/analyser.go
```diff
@@ -8,14 +8,255 @@ import (
...
+	}
+
+	if schema.IsPrimitiveType(to) {
+		switch to {
+		case schema.AnyType:
+			return okIf(true)
+		case schema.NumberType, schema.IntType:
+			return okIf(from == schema.NumberType || from == schema.IntType)
+		case schema.StringType:
+			// Resources can coerce into strings (by implicitly calling urn)
```
AaronFriel, line 139: Where does this implicit conversion occur?

pkg/pulumiyaml/analyser.go
```diff
@@ -8,14 +8,255 @@ import (
...
+			_, isResource := from.(*schema.ResourceType)
+			return okIf(isResource || from == schema.StringType ||
+				// Since we don't have a fn(number) -> string function, we coerce numbers to strings
```
AaronFriel, line 142: Also confused on this.

pkg/pulumiyaml/analyser.go
```diff
@@ -8,14 +8,255 @@ import (
...
+	case *schema.TokenType:
+		if to.UnderlyingType != nil {
+			return isAssignable(from, to.UnderlyingType)
+		}
+		return &notAssignable{reason: fmt.Sprintf("Unknown opaque type: %s", to.Token), internal: true}
+	default:
+		// We mark this an internal error since we don't recognize the type.
+		// This means we are missing a type, not that the user has an invalid
+		// program.
+		return &notAssignable{reason: fmt.Sprintf("Unknown type: %[1]s (%[1]T)", to), internal: true}
```
AaronFriel, line 237: I think once we're satisfied with this, I'd like to see a PR moving this logic to pulumi/pulumi and/or using a tool like https://github.com/BurntSushi/go-sumtype to add a lint pass to ensure the case statement here is exhaustive.

pkg/pulumiyaml/analyser.go
```diff
@@ -8,14 +8,255 @@ import (
...
+			if notAssignable != nil {
+				return fail.Because(notAssignable).Property(prop.Name)
+			}
+		}
+		return nil
+	case *schema.TokenType:
+		if to.UnderlyingType != nil {
+			return isAssignable(from, to.UnderlyingType)
+		}
+		return &notAssignable{reason: fmt.Sprintf("Unknown opaque type: %s", to.Token), internal: true}
```
AaronFriel, line 232: These errors look good, but I'd like to see a test case for the diagnostics output.

pkg/pulumiyaml/analyser.go
```diff
@@ -8,14 +8,255 @@ import (
...
+		if !ok {
+			return fail
+		}
+		return okIfAssignable(isAssignable(from.ElementType, to.ElementType))
+	case *schema.MapType:
+		switch from := from.(type) {
+		case *schema.MapType:
+			return okIfAssignable(isAssignable(from.ElementType, to.ElementType))
+		case *schema.ObjectType:
+			// YAML does not distinguish between maps and arrays, but our type system does.
```
AaronFriel, line 177: nit: Maps and objects?

pkg/pulumiyaml/analyser.go
```diff
@@ -157,23 +417,344 @@ func (tc *typeCache) typeInvoke(ctx *evalContext, t *ast.InvokeExpr) bool {
...
+	case *ast.PropertyName:
+		properties := map[string]schema.Type{}
+		switch root := codegen.UnwrapType(root).(type) {
+		case *schema.ObjectType:
+			for _, prop := range root.Properties {
+				properties[prop.Name] = prop.Type
+			}
+		case *schema.ResourceType:
+			for _, prop := range root.Resource.Properties {
+				properties[prop.Name] = prop.Type
```
AaronFriel, line 528: We don't have to handle optionals here, right?

pkg/pulumiyaml/analyser.go
iwahbe: I'm not 100% sure where the conversion is in the codebase (it might be at the RPC level), but I can do
```yaml
resources:
  my-bucket:
    type: aws:s3/bucket:Bucket
    properties:
      website:
        indexDocument: index.html
  index.html:
    type: aws:s3/bucketObject:BucketObject
    properties:
      bucket: ${my-bucket}
      source:
        Fn::StringAsset: <h1>Hello, world!</h1>
```

and [`aws:s3/bucketObject:BucketObject.bucket`](https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/schema.json#L314298-L314302) is of type string.

AaronFriel (commented):
Mostly looks good, curious about those coercion rules though.

pkg/pulumiyaml/analyser.go
iwahbe: I am attempting to preserve the current behavior. Consider the following example, taken from the `aws-fargate-pp` example. Consider this snippet:

```yaml
  ...
  appTask:
    type: aws:ecs:TaskDefinition
    properties:
      family: fargate-task-definition
      cpu: 256
      memory: 512
  ...
```

Here both [`cpu`](https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/schema.json#L256979-L256983) and [`memory`](https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/schema.json#L257006-L257010) take strings, but are given as numbers. This just works. I didn't want to break existing code.

iwahbe: Also of note, if we image a Pulumi YAML program with an important number that is refactored to a variable and then given to properties of number and string types:
```yaml
variables:
  myNumber: 42
resources:
  handleNumber:
    asNumberType: ${myNumber}
    asStringType: ${myNumber}
```

I feel like we want this to just work. Other languages would have us call `{ asNumberType: myNumber, asStringType: myNumber.string() }` but we don't have that support in Pulumi YAML. We could require that the user calls `Fn::ToJSON`, but I don't think we want to require that.

pkg/pulumiyaml/analyser.go
iwahbe: I would be happy to add a sumtype lint to this. I don't think the logic should be moved into `pulumi/pulumi` because the coercion rules are specific to Pulumi YAML. We are writing the typing rules of Pulumi YAML and also describing how Pulumi YAML types map onto pulumi types. This shouldn't apply to any other language, since we will be using the type system of the host language.

pkg/pulumiyaml/analyser.go
iwahbe: We have to consider them, as we do on line [521](https://github.com/pulumi/pulumi-yaml/blob/iwahbe/type-checker/pkg/pulumiyaml/analyser.go#L521). I'm allowing people to assume that optional properties exist. I'm not 100% sure that I am answering your question. Let me know if I misunderstood.

pkg/pulumiyaml/analyser.go
iwahbe: Check out https://github.com/pulumi/pulumi-yaml/pull/228/commits/2f7005f9ce139ae7933c4413974551c155d8bfc0 and https://github.com/pulumi/pulumi-yaml/pull/228/commits/5871d69f6821ec09b869a1a5476f68f0d6ebdc4f.

pkg/pulumiyaml/analyser.go
AaronFriel: That explanation makes perfect sense.

pkg/pulumiyaml/analyser.go
AaronFriel: C'est la vie. Thanks for looking!

pkg/pulumiyaml/analyser.go
AaronFriel: I was pattern matching against this code from the binder, where non-required props have their type wrapped in `OptionalType{}`:
 https://github.com/pulumi/pulumi/blob/36cbf572f45790f6859a8da3f8d67f6b32620fca/pkg/codegen/pcl/binder_schema.go#L273-L280

AaronFriel (approved):

pkg/pulumiyaml/analyser.go
iwahbe: I don't think we need to do anything like that. I'm actually very skeptical of that piece of code, since [this](https://github.com/pulumi/pulumi/blob/36cbf572f45790f6859a8da3f8d67f6b32620fca/pkg/codegen/schema/schema.go#L362-L365) is the definition of `IsRequired`. 

It means that we are doing this:
```go
		for _, prop := range src.Resource.Properties {
			typ := prop.Type
                        
			if _, optional := typ.(*schema.OptionalType); optional {
				typ = &schema.OptionalType{ElementType: typ}
			}


			properties[prop.Name] = b.schemaTypeToTypeOrConst(typ, prop)
		}
```

AaronFriel: Hmm, worth checking if we can simplify there, then.

PR #230, opened 2022-05-23
[exec] Consume PackageReferences

pgavlin:
Change the package loader to use schema.PackageReference values rather
than *schema.Package values. The former load types, resources, and
functions as they are needed rather than up-front, which can drastically
reduce the peak memory consumption of the executor.


AaronFriel:
Rebased on top of Ian's recent changed and pushed back.

iwahbe (approved):
LGTM

PR #231, opened 2022-05-24
Remove mutexes introduced due to concurrency bug

AaronFriel:
Blocked on resolution of & PR:
- https://github.com/pulumi/pulumi/issues/9667
- https://github.com/pulumi/pulumi/pull/9675


AaronFriel:
Updating go.mod to target the branch for now so that this is reviewable.

iwahbe (approved):

AaronFriel:
Updating go.mod to target upstream after this fix was merged.

joeduffy:
Totally random, but in cases like this, I always love when it’s possible to “assert” the safety of the concurrent access. Many times when removing a mutex lock, that means you can assert a certain lock is held as a precondition to the method call. Is that sort of thing relevant or possible here?

PR #237, opened 2022-05-26
Enable GitHub "CodeQL" analysis as an experiment

AaronFriel:
This workflow was added as-is from from the "Security" tab of the repo.


iwahbe (commented):
Will we actually follow these recommendations? We get a major security vulnerability reported now, it is not actually correct. It's warning that `make([]int, len(b)+1)` could over allocate, which isn't something I am worried about.

P.S. It will slow down the PR workflow by about 20% (from this run).

AaronFriel:
Agreed, doesn't seem worth it based on the results here.

PR #241, opened 2022-06-01
[expr] Allow expressions as the second arg to Fn::Join

### LINKED ISSUE #240: Support  in Fn::Join
In YAML like the example below, an error is thrown that the second argument must be a list. This isn't true, we can relax it like we have other functions to allow any expression, so long as that expression evaluates to a list.

```yaml    
variables:
  certificateChainPem: 
    Fn::Join:
      - "\n"
      - ${certificate.chainCertPems}
```

Error: `the second argument to Fn::Join must be a list; the second argument to Fn::Join must be a list`

AaronFriel:
Fixes #240

### DIFF
--- a/pkg/pulumiyaml/run_test.go
+++ b/pkg/pulumiyaml/run_test.go
@@ -1607,6 +1607,10 @@ variables:
     Fn::Join:
       - "-"
       - ${inputs}
+  foo-err:
+    Fn::Join:
+      - "-"
+      - ${inputs[1]}
 `
 
 	tmpl := yamlTemplate(t, strings.TrimSpace(text))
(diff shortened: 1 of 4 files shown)


pkg/pulumiyaml/run_test.go
```diff
@@ -1565,6 +1565,67 @@ outputs:
...
+	var diagStrings []string
+	for _, v := range diags {
+		diagStrings = append(diagStrings, diagString(v))
+	}
+	assert.ElementsMatch(t, diagStrings,
+		[]string{
+			"<stdin>:12:9: the second argument to Fn::Join must be a list of strings, found a number at index 0",
+			"<stdin>:12:9: the second argument to Fn::Join must be a list of strings, found an object at index 1",
+			"<stdin>:12:9: the second argument to Fn::Join must be a list of strings, found a list at index 2",
+			"<stdin>:12:9: the second argument to Fn::Join must be a list of strings, found a boolean at index 3",
```
AaronFriel, line 1624: Is there a way to neatly print the _value_ that was the issue? And should we stop at a certain number of errors?

Both of these would be good follow-up enhancements to the diagnostic, I think.

iwahbe (approved):
I'm going to improve this, but we need to check that the second argument is a list before merging. As written [we cast it](https://github.com/pulumi/pulumi-yaml/blob/6bed2cbac31fd6b63b5510b4e1abacc2d1881670/pkg/pulumiyaml/run.go#L1334), which will panic.

AaronFriel:
@iwahbe that's an easy case to handle & test to add to prevent a panic.

Issue #235, opened 2022-05-25
Fn::ToBase64 isn't in docs


dixler:
### What happened?

`Fn::ToBase64` isn't in the [README.md](https://github.com/pulumi/pulumi-yaml#built-in-functions)

### Steps to reproduce

[README.md](https://github.com/pulumi/pulumi-yaml#built-in-functions)

### Expected Behavior

It to be in the docs.

### Actual Behavior

It's not in the docs.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

RobbieMcKinstry:
I'll hop on this one shortly!

PR #245, opened 2022-06-03
Document Fn::ToBase64

RobbieMcKinstry:
This PR closes https://github.com/pulumi/pulumi-yaml/issues/235 by providing documentation for the built-in `Fn::Base64` function.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,9 +22,6 @@
 
 ### Bug Fixes
 
-- Document Fn::ToBase64 in README.
-  [#245](https://github.com/pulumi/pulumi-yaml/pull/245)
-
 - De-duplicate error message added during pre-eval checking.
   [#207](https://github.com/pulumi/pulumi-yaml/pull/207)
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,6 +22,9 @@
 
 ### Bug Fixes
 
+- Document Fn::ToBase64 in README.
+  [#245](https://github.com/pulumi/pulumi-yaml/pull/245)
+
 - De-duplicate error message added during pre-eval checking.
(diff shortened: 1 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -22,6 +22,9 @@
 
 ### Bug Fixes
 
+- Document Fn::ToBase64 in README.
+  [#245](https://github.com/pulumi/pulumi-yaml/pull/245)
```
mikhailshilkov, line 26: IMO you don't need to include this to README as this is not a product change, just apply the `no-changelog-required` label.

README.md
```diff
@@ -271,6 +271,18 @@ The last two lines are equivalent as if the variable were substituted for its va
 
 In any expression location, an object containing a single key beginning with "Fn::" calls a built-in function.
 
+##### `Fn::ToBase64`
+
+This function encodes a UTF-8 string with the base64-encoding using the [standard base-64 encoding](https://pkg.go.dev/encoding/base64#pkg-variables). **This will fail if the result is not a valid UTF-8 string**
```
mikhailshilkov, line 276: When is it possible that the result is not a valid base64 string?

iwahbe, line 276:
```suggestion
Encodes a UTF-8 string with Base64 encoding using the [standard Base64 encoding](https://pkg.go.dev/encoding/base64#pkg-variables). **This will fail if the result is not a valid UTF-8 string**
```

I'm looking at the way other function descriptions are written. They all start with a verb ("Converts", "Calls"...). In the `Fn::FromBase64` description we write "Base64" instead of "base64".

RobbieMcKinstry: Whoops, that was my carelessness with copy-paste!

RobbieMcKinstry: I addressed this comment in a new commit. Thanks! Good catch!

iwahbe (approved):
LGTM

mikhailshilkov (approved):
LGTM, thank you

Issue #252, opened 2022-06-16
Clunky error message when misusing dependsOn


joeduffy:
I got a weird error message if I put `dependsOn` in the wrong place. It seems to be just printing the contents of a struct of some sort, although I'm not sure which. Am guessing this could be a bug?

```
    error: Resource command:local:Command has invalid key 'dependsOn'

      on Pulumi.yaml line 190:
     190:     dependsOn:

    'dependsOn' exists under options, e.g.

    playAnsiblePlaybookCmd:
      # ...
      options:
        dependsOn: &{{{%!s(*yaml.Node=&{2 0 !!seq   <nil> [0xc0006de3c0 0xc0006de460]    191 7}) %!s(*hcl.Range=&{Pulumi.yaml {191 7 0} {192 24 0}}) <nil>}} [%!s(*syntax.StringNode=&{{{0xc0006de3c0 0xc0006e38c0 renderPlaybookCmd}} renderPlaybookCmd}) %!s(*syntax.StringNode=&{{{0xc0006de460 0xc0006e3900 updatePythonCmd}} updatePythonCmd})]}
```

PR #257, opened 2022-06-18
Require fmt.Stringer for a Node

iwahbe:
Fixes: #252

### DIFF
--- a/pkg/pulumiyaml/syntax/nodes.go
+++ b/pkg/pulumiyaml/syntax/nodes.go
@@ -2,10 +2,18 @@
 
 package syntax
 
-import "github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


pkg/pulumiyaml/syntax/nodes.go
```diff
@@ -141,6 +164,14 @@ func (n *ListNode) Index(i int) Node {
 	return n.elements[i]
 }
 
+func (n *ListNode) String() string {
+	s := make([]string, len(n.elements))
+	for i, v := range n.elements {
+		s[i] = v.String()
+	}
+	return fmt.Sprintf("[ %s ]", strings.Join(s, ", "))
```
AaronFriel, line 172: Could you add a test that shows what these types will look like in practice?

AaronFriel (commented):
I accidentally left a "single comment" outside of this review and GitHub requires that I write something to submit a review. :)

AaronFriel (approved):

Issue #254, opened 2022-06-16
Incorrect error message with malformed Fn::Invoke


joeduffy:
Oddly, if I have the following:

```yaml
name: pulumi-ansible-wordpress
runtime: yaml
description: Deploy an EC2 Wordpress server using Pulumi and Ansible.

configuration:
  publicKeyPath:
    type: String
    required: true
  privateKeyPath:
    type: String
    required: true
  dbInstanceSize:
    type: String
    default: db.t3.small
  dbName:
    type: String
    default: wordpressdb
  dbUsername:
    type: String
    default: admin
  dbPassword:
    type: String
    required: true
  ec2InstanceSize:
    type: String
    default: t3.small

variables:
  # Dynamically fetch AZs so we can spread across them.
  availabilityZones:
    Fn::Invoke:
      Function: aws:getAvailabilityZones
  # Dynamically query for the Amazon Linux AMI in this region.
  awsLinuxAmi:
    Fn::Invoke:
      function: aws:ec2:getAmi
      arguments:
        filters:
          - name: name
            values: ["amzn-ami-hvm-*-x86_64-ebs"]
        owners: ["137112412989"] # Amazon
        mostRecent: true

...
```

I get the following error: `error: failed to discover plugin requirements: Pulumi.yaml:8,5-13: Field 'required' does not exist on Object 'configuration.publicKeyPath'; Existing fields are: 'secret', 'type', 'default', and 3 other diagnostic(s)`.

The thing is this seems to have nothing to do with the real problem. The problem is that the second `Fn::Invoke` uses lowercase `function` and `argument` properties. They need to be uppercase, as in `Function` and `Argument`.

Indeed, fixing that makes the seemingly unrelated error go away!

mikhailshilkov:
As mentioned in https://github.com/pulumi/pulumi/issues/9890, `required` is not a valid property, so the error is legit. You can also see `and 3 other diagnostic(s)` which hides your actual error. Once I fix the required warnings, I get `could not load yaml program: failed to load the template: Pulumi.yaml:35,7-41,25: missing function name ('Function'); missing function name ('Function')`.

I suppose we can use this issue to track improvements to error presentations. Why do we need to hide the warnings under "and 3 other diagnostic(s)"? Can we sort the diagnostics so that errors are ahead of warnings?

AaronFriel:
I think after landing the fix to #253 in PR #258, we should show all diagnostics.

PR #279, opened 2022-07-12
Show all errors/ warning messages

aq17:
Addresses #254


aq17:
Example output:
```
error: failed to discover plugin requirements: 
-error: Pulumi.yaml:13,7-41: missing function name ('Function'); missing function name ('Function')
-warning: Pulumi.yaml:8,5-13: Field 'required' does not exist on Object 'configuration.name'; Existing fields are: 'secret', 'type', 'default'
```

AaronFriel:
@aq17 the extended output looks good, is it easy to remove the duplication on the error though?

```
-error: Pulumi.yaml:13,7-41: missing function name ('Function'); missing function name ('Function')
```

AaronFriel:
I think tests are failing due to #280, which might be due to merging a PR that bumped codegen from pulumi/pulumi without updating tests. Once #281 is merged, I expect the test failures to disappear.

aq17:
> @aq17 the extended output looks good, is it easy to remove the duplication on the error though?
> 
> ```
> -error: Pulumi.yaml:13,7-41: missing function name ('Function'); missing function name ('Function')
> ```

The `hcl` pkg we import from Hashicorp prints error/ warning messages with this format `return fmt.Sprintf("%s: %s; %s", d.Subject, d.Summary, d.Detail)`; right now we set Detail as Summary if there are no additional details, so we could omit but it'd still have the semicolon

```
error: failed to discover plugin requirements: 
-error: Pulumi.yaml:13,7-41: missing function name ('Function'); 
-warning: Pulumi.yaml:8,5-13: Field 'required' does not exist on Object 'configuration.name'; Existing fields are: 'secret', 'type', 'default'
```
Does that work?

AaronFriel:
Yeah, I think that looks better to me - let's omit the duplication there.

AaronFriel:
@iwahbe curious if you agree here on removing the "detail" if it's duplicative. I think it makes sense?

iwahbe:
I'm happy to remove the duplicative detail.

AaronFriel (approved):

AaronFriel (changes requested):
Tests need to be updated with the changed error reporting:

https://github.com/pulumi/pulumi-yaml/runs/7311777561?check_suite_focus=true#step:13:770

PR #284, opened 2022-07-13
Support options on Fn::Invoke

aq17:
`Fn::Invoke` needs the ability to specify  InvokeOptions (`Parent, Provider, Version, PluginDownloadUrl`) to use for the function invocation

### DIFF
--- a/pkg/pulumiyaml/expr.go
+++ b/pkg/pulumiyaml/expr.go
@@ -60,6 +60,7 @@ func getExpressionDependencies(deps *[]*ast.StringExpr, x ast.Expr) {
 		sx := ast.StringSyntax(syntax.StringSyntax(x.Syntax().Syntax(), name))
 		*deps = append(*deps, sx)
 	case *ast.InvokeExpr:
+		getExpressionDependencies(deps, x.Args())
 		if x.CallOpts.Parent != nil {
 			getExpressionDependencies(deps, x.CallOpts.Parent)
 		}
--- a/pkg/pulumiyaml/expr.go
+++ b/pkg/pulumiyaml/expr.go
@@ -59,8 +59,6 @@ func getExpressionDependencies(deps *[]*ast.StringExpr, x ast.Expr) {
 		name := x.Property.RootName()
 		sx := ast.StringSyntax(syntax.StringSyntax(x.Syntax().Syntax(), name))
 		*deps = append(*deps, sx)
-	case ast.BuiltinExpr:
-		getExpressionDependencies(deps, x.Args())
 	case *ast.InvokeExpr:
 		if x.CallOpts.Parent != nil {
 			getExpressionDependencies(deps, x.CallOpts.Parent)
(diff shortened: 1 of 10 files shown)


aq17:
@iwahbe @AaronFriel I'm not convinced I tested this thoroughly; my program worked with 
```
variables:
  testing:
    Fn::Invoke:
      Function: aws:getCallerIdentity
      Options:
        version: v5.9.0
      Return: arn
```
but I think trying to use a function in a specified provider that isn't included in the default provider yet would be better – any suggestions?

aq17:
@AaronFriel thanks for walking me through this, tested it out with a simple example and looks good

```
variables:
  actualRegion:
    Fn::Invoke:
      Function: aws:getRegion
      Options:
        Providers: ${myProvider}
      Return: name

resources:
  myProvider:
    type: pulumi:providers:aws
    properties:
      region: us-east-2
     
outputs: 
  expectedRegion: ${myProvider.region}
  actualRegion: ${actualRegion}
```
`actualRegion` is `us-east-1` if the provider is not specified, `us-east-2` if it is.

iwahbe (commented):
Looks good. I have two comments:
1. How does this interact with [`sort.getExpressionDependencies`](https://github.com/pulumi/pulumi-yaml/blob/61b24a73cba591d3b83d38e0893f1e0a5be56d03/pkg/pulumiyaml/expr.go#L39)? I think you will need to special case `ast.InvokeExpr` similar to `ast.ResourceDecl`.
2. Can you add a test?

pkg/pulumiyaml/expr.go
```diff
@@ -61,6 +61,12 @@ func getExpressionDependencies(deps *[]*ast.StringExpr, x ast.Expr) {
...
 	case ast.BuiltinExpr:
 		getExpressionDependencies(deps, x.Args())
-
+	case *ast.InvokeExpr:
+		if x.CallOpts.Parent != nil {
+			getExpressionDependencies(deps, x.CallOpts.Parent)
+		}
+		if x.CallOpts.Providers != nil {
+			getExpressionDependencies(deps, x.CallOpts.Providers)
+		}
```
aq17, line 70: @iwahbe believe this is what you meant, parsing the parent and providers resource dependencies

iwahbe (commented):
This looks good! I think the only system missing is type-checking. That would live in https://github.com/pulumi/pulumi-yaml/blob/f3a244e33f5687e93593a5f3c7c3e99ae553dd28/pkg/pulumiyaml/analyser.go#L557
I appreciate the test, but it doesn't look like it actually confirms that the invoke was called with the `Providers` option. I don't see where we assert that `test:invoke:type` was passed the provider.

aq17:
> This looks good! I think the only system missing is type-checking. That would live in
> 
> https://github.com/pulumi/pulumi-yaml/blob/f3a244e33f5687e93593a5f3c7c3e99ae553dd28/pkg/pulumiyaml/analyser.go#L557
> 
> 
> I appreciate the test, but it doesn't look like it actually confirms that the invoke was called with the `Providers` option. I don't see where we assert that `test:invoke:type` was passed the provider.

I added a new resource `test:invoke:type2` so we wouldn't have to pass in a Provider for the other examples

iwahbe (approved):
This looks good to me. LGTM

PR #286, opened 2022-07-19
Update pulumi dependency

aq17:
Update pulumi dependency to support `convert` for stack references

Corresponding change in `pulumi/pulumi` merged: https://github.com/pulumi/pulumi/pull/10174

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,8 +12,8 @@ require (
 	github.com/hexops/autogold v1.3.0
 	github.com/iancoleman/strcase v0.2.0
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/pkg/v3 v3.35.3
-	github.com/pulumi/pulumi/sdk/v3 v3.35.3
+	github.com/pulumi/pulumi/pkg/v3 v3.36.1-0.20220719024731-e020a12b4470
+	github.com/pulumi/pulumi/sdk/v3 v3.36.1-0.20220719024731-e020a12b4470
 	github.com/stretchr/testify v1.7.1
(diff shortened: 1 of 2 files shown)


iwahbe:
It looks like we broke some codegen tests.

aq17:
> It looks like we broke some codegen tests.

@AaronFriel @iwahbe I think this may be due to the circular link

AaronFriel:
@aq17 there could be other changes as I've been updating and improving codegen.

What does the diff look like if you run `go test -timeout 5m -run ^TestGenerateExamples$ github.com/pulumi/pulumi-yaml/pkg/tests -v`.

aq17:
> @aq17 there could be other changes as I've been updating and improving codegen.
> 
> What does the diff look like if you run `go test -timeout 5m -run ^TestGenerateExamples$ github.com/pulumi/pulumi-yaml/pkg/tests -v`.

Lot of errors `unable to resolve function name: internal error loading package "aws"/"azure-native"` because `.json` test files are missing locally cause of the .gitignore

iwahbe:
>@AaronFriel Lot of errors unable to resolve function name: internal error loading package "aws"/"azure-native" because .json test files are missing locally cause of the .gitignore

I don't understand how this is because of the circular dependency.

aq17:
> > @AaronFriel Lot of errors unable to resolve function name: internal error loading package "aws"/"azure-native" because .json test files are missing locally cause of the .gitignore
> 
> I don't understand how this is because of the circular dependency.

These errors are not, I meant for the codegen failures?

PR #294, opened 2022-07-26
Add sugar for `Fn::Invoke`

### LINKED ISSUE #288: Consider sugar for `Fn::Invoke`
## Hello!

## Issue details

We would like to make it more ergonomic to call invokes. We should consider adding the following sugar:
```yaml
Fn::pkg:module:Name:
  Arg1: ${Arg1}
  Arg2: ${Arg2}
# Would be equivalent to:
Fn::Invoke:
  Function: pkg:module:Name
  Arguments:
    Arg1: ${Arg1}
    Arg2: ${Arg2}
```

This will make invokes similar in calling convention to our built-in functions. 

Upsides:
- Unifies invoke and builtin fn syntax
- Simplifies syntax, loosing a level of indentation

Downsides:
- We will still need to support the `Fn::Invoke` syntax, since it is necessary for supplying invoke options.
- More ways of doing things and more surface area for us.

aq17:
Fixes #288 
Supports a more natural way of calling `Fn::Invoke` for calls with only `Arguments`, i.e.
```
Fn::pkg:module:Name:
  Arg1: ${Arg1}
  Arg2: ${Arg2}
# Would be equivalent to:
Fn::Invoke:
  Function: pkg:module:Name
  Arguments:
    Arg1: ${Arg1}
    Arg2: ${Arg2}
```

`Fn::Invoke` will still be supported for supplying invoke `Options` and `Return` values 

Todo: update docs

### DIFF
--- a/examples/azure-container-apps/Pulumi.yaml
+++ b/examples/azure-container-apps/Pulumi.yaml
@@ -20,7 +20,7 @@ variables:
       registryName: ${registry.name}
   adminUsername: ${adminRegistryCreds.username}
   adminPasswords:
-    Fn::Secret: ${adminRegistryCreds.password}
+    Fn::Secret: ${adminRegistryCreds.passwords}
 resources:
   resourceGroup:
     type: azure-native:resources:ResourceGroup
@@ -70,8 +70,8 @@ resources:
       kubeEnvironmentId: ${kubeEnv}
       configuration:
         ingress:
-            external: true
-            targetPort: 80
+          external: true
+          targetPort: 80
         registries:
           - server: ${registry.loginServer}
             username: ${adminUsername}
--- a/examples/azure-container-apps/Pulumi.yaml
+++ b/examples/azure-container-apps/Pulumi.yaml
@@ -14,13 +14,12 @@ variables:
           resourceGroupName: ${resourceGroup.name}
           workspaceName: ${workspace.name}
         Return: primarySharedKey
-  adminUsername:
-    Fn::Invoke:
-      Function: azure-native:containerregistry:listRegistryCredentials
+  adminRegistryCreds:
+    Fn::azure-native:containerregistry:listRegistryCredentials:
       Arguments:
         resourceGroupName: ${resourceGroup.name}
         registryName: ${registry.name}
-      Return: username
+  adminUsername: ${adminRegistryCreds.username}
   adminPasswords:
     Fn::Secret:
       Fn::Invoke:
--- a/examples/azure-container-apps/Pulumi.yaml
+++ b/examples/azure-container-apps/Pulumi.yaml
@@ -20,13 +20,7 @@ variables:
       registryName: ${registry.name}
   adminUsername: ${adminRegistryCreds.username}
   adminPasswords:
-    Fn::Secret:
-      Fn::Invoke:
-        Function: azure-native:containerregistry:listRegistryCredentials
-        Arguments:
-          resourceGroupName: ${resourceGroup.name}
-          registryName: ${registry.name}
-        Return: passwords
+    Fn::Secret: ${adminRegistryCreds.password}
 resources:
   resourceGroup:
     type: azure-native:resources:ResourceGroup
--- a/examples/azure-container-apps/Pulumi.yaml
+++ b/examples/azure-container-apps/Pulumi.yaml
@@ -16,9 +16,8 @@ variables:
         Return: primarySharedKey
   adminRegistryCreds:
     Fn::azure-native:containerregistry:listRegistryCredentials:
-      Arguments:
-        resourceGroupName: ${resourceGroup.name}
-        registryName: ${registry.name}
+      resourceGroupName: ${resourceGroup.name}
+      registryName: ${registry.name}
   adminUsername: ${adminRegistryCreds.username}
   adminPasswords:
     Fn::Secret:
--- a/pkg/pulumiyaml/ast/expr.go
+++ b/pkg/pulumiyaml/ast/expr.go
@@ -747,7 +747,7 @@ func tryParseFunction(node *syntax.ObjectNode) (Expr, syntax.Diagnostics, bool) 
 		// Fn::Invoke can be called as Fn::{pkg}:{Name}
 		if match, _ := regexp.MatchString(fnInvokeRegexPattern, k); match {
 			// transform the node into standard Fn::Invoke format
-			fnVal := strings.Split(k, "Fn::")[1]
+			fnVal := strings.TrimPrefix(k, "Fn::")
 			if _, ok := kvp.Value.(*syntax.ObjectNode); ok {
 				kvp.Value = syntax.Object(
 					syntax.ObjectPropertyDef{
--- a/pkg/pulumiyaml/ast/expr.go
+++ b/pkg/pulumiyaml/ast/expr.go
@@ -744,7 +744,8 @@ func tryParseFunction(node *syntax.ObjectNode) (Expr, syntax.Diagnostics, bool) 
 	default:
 		var diags syntax.Diagnostics
 		k := kvp.Key.Value()
-		// Fn::Invoke can be called as Fn::{pkg}:{Name}
+		// Fn::Invoke can be called as FN::${pkg}:${module}(:${name})?
+		// error is thrown if regex pattern cannot be parsed — handled by `regex.MustCompile(fnInvokeRegexPattern)`
 		if match, _ := regexp.MatchString(fnInvokeRegexPattern, k); match {
 			// transform the node into standard Fn::Invoke format
 			fnVal := strings.TrimPrefix(k, "Fn::")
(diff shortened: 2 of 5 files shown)


aq17:
Still need to
- add test
- update docs
but curious if this approach of manipulating the node into the standard `Fn::Invoke` structure looks reasonable @iwahbe @AaronFriel

AaronFriel (commented):
The regex splitting is applying a bit too much power to the test, I think it's OK to forbid `Fn::` prefixed keys in maps for now (we'll address later if it's an issue).

I'd like to see one or more of our examples updated. I think this is a good candidate because it's unnecessarily long: we're calling this invoke twice to get two different values. The diff here will be instructive of what it looks like to use the shorthand.

https://github.com/pulumi/pulumi-yaml/blob/6d5979e8b471a986ebd6e1707cb94c785e10e157/examples/azure-container-apps/Pulumi.yaml#L17-L31

aq17:
> The regex splitting is applying a bit too much power to the test, I think it's OK to forbid `Fn::` prefixed keys in maps for now (we'll address later if it's an issue).
> 
> I'd like to see one or more of our examples updated. I think this is a good candidate because it's unnecessarily long: we're calling this invoke twice to get two different values. The diff here will be instructive of what it looks like to use the shorthand.
> 
> https://github.com/pulumi/pulumi-yaml/blob/6d5979e8b471a986ebd6e1707cb94c785e10e157/examples/azure-container-apps/Pulumi.yaml#L17-L31

Do we still want to support `Return` with the shorthand syntax? I'm not quite sure that will work given it's a top-level `Fn::pkg:module:name` and then only a map of arguments beneath, as Ian outlined in the [issue example ](https://github.com/pulumi/pulumi-yaml/issues/288)

pkg/pulumiyaml/ast/expr.go
```diff
@@ -4,13 +4,16 @@ package ast
...
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/syntax"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
+const fnInvokeRegexPattern = "Fn::([A-Za-z]+):([A-Za-z]+)(:)?([A-Za-z]*)"
```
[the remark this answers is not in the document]
aq17: If we assume anything prefixed with `Fn::` that isn't one of our documented function calls intends to use the shorthand for Invoke, then we'll just split on `Fn` and let `parseInvoke` handle any errors, removing our [current default (error) case](https://github.com/pulumi/pulumi-yaml/blob/main/pkg/pulumiyaml/ast/expr.go#L741-L750)?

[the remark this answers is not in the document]
iwahbe: Anything of the form `Fn::${TEXT}` is a built in function call. If it doesn't match an existing built in function call, we should error assuming that is what the user was trying to type. 

Anything of the form `FN::${TEXT}:${TEXT}(:${TEXT})?` is a sugared invoke. 

It is already a warning to have map keys that start with `Fn::`.

aq17 (commented):
Does updating the example to show when one might want to use `Fn::Invoke` (to `Return` a specific attribute for secrets) versus using the simplified syntax and then defining additional variables (such as `adminUsername`) for individual attributes makes sense? @iwahbe @AaronFriel

iwahbe:
I think the example you updated makes sense. I think it shows improvement in the code, but the real impact will be for utility functions defined in providers:

```yaml
result:
  Fn::Invoke:
    Function: str:regex:Split
    Arguments:
      string: foo---bar
      split: -+
    Return: result
first: ${result[0]}
```
vs
```yaml
first:
  Fn::str:regex:Split:
    string: foo---bar
    split: -+
first: ${strSplit.result[0]}
```

I don't think we have an example of that right now though.

examples/azure-container-apps/Pulumi.yaml
```diff
@@ -14,13 +14,11 @@ variables:
...
-      Function: azure-native:containerregistry:listRegistryCredentials
-      Arguments:
-        resourceGroupName: ${resourceGroup.name}
-        registryName: ${registry.name}
-      Return: username
+  adminRegistryCreds:
+    Fn::azure-native:containerregistry:listRegistryCredentials:
+      resourceGroupName: ${resourceGroup.name}
+      registryName: ${registry.name}
+  adminUsername: ${adminRegistryCreds.username}
```
iwahbe, line 21: Can we replace `adminPasswords` with `Fn::Secret: ${adminRegistryCreds.password}`?

pkg/pulumiyaml/ast/expr.go
```diff
@@ -741,7 +744,32 @@ func tryParseFunction(node *syntax.ObjectNode) (Expr, syntax.Diagnostics, bool)
 	default:
 		var diags syntax.Diagnostics
 		k := kvp.Key.Value()
-		if strings.HasPrefix(k, "Fn::") {
+		// Fn::Invoke can be called as Fn::{pkg}:{Name}
+		if match, _ := regexp.MatchString(fnInvokeRegexPattern, k); match {
+			// transform the node into standard Fn::Invoke format
+			fnVal := strings.Split(k, "Fn::")[1]
```
iwahbe, line 750:
```suggestion
			fnVal := strings.TrimPrefix(k, "Fn::")
```

pkg/pulumiyaml/ast/expr.go
```diff
@@ -741,7 +744,32 @@ func tryParseFunction(node *syntax.ObjectNode) (Expr, syntax.Diagnostics, bool)
 	default:
 		var diags syntax.Diagnostics
 		k := kvp.Key.Value()
-		if strings.HasPrefix(k, "Fn::") {
+		// Fn::Invoke can be called as Fn::{pkg}:{Name}
+		if match, _ := regexp.MatchString(fnInvokeRegexPattern, k); match {
```
iwahbe, line 748: Can we address the returned error here? We should at least leave a comment saying what the error will be.

pkg/pulumiyaml/ast/expr.go
```diff
@@ -4,13 +4,16 @@ package ast
...
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/syntax"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
+const fnInvokeRegexPattern = "Fn::.+:.+(:.+)?"
```
iwahbe, line 15: We try to avoid compiling precomputed regexes in functions to avoid recompiling them. This is idiomatic:
```suggestion
var fnInvokeRegex = regex.MustCompile("Fn::.+:.+(:.+)?")
```

pkg/pulumiyaml/run_invoke_test.go
```diff
@@ -143,6 +143,72 @@ runtime: yaml
...
+	requireNoErrors(t, tmpl, diags)
+}
+
+func TestInvokeNoInputsSugar(t *testing.T) {
+	t.Parallel()
+
+	const text = `
+variables:
+  config:
+    Fn::test:invoke:empty:
```
iwahbe, line 200: I'm surprised that this is valid. 😀 I would have expected that you needed something like `Fn::test:invoke:empty: {}`

pkg/pulumiyaml/run_invoke_test.go
```diff
@@ -143,6 +143,72 @@ runtime: yaml
 	requireNoErrors(t, tmpl, diags)
 }
 
+func TestInvokeVariableSugar(t *testing.T) {
```
iwahbe, line 146: Should we test on parsing instead of execution? Since this is sugar, we don't really need a test on the execution of the invoke. 

CC @AaronFriel

pkg/pulumiyaml/ast/expr.go
aq17: I believe the error will be if our regex pattern can't be parsed, which should be handled by `regex.MustCompile(fnInvokeRegexPattern)` – I'll add a comment

pkg/pulumiyaml/run_invoke_test.go
aq17: Don't think we have to modify to error if `{}` isn't provided, but it looks nicer so we should probably use it in future examples

pkg/pulumiyaml/run_invoke_test.go
AaronFriel: We could, but I think this test is fine. We've made it really simple in this library to write end to end tests.

aq17:
The failing lint tests seem to be an anomaly, ran `gofmt` and no changes

AaronFriel (approved):
lgtm

PR #295, opened 2022-07-26
fix: update language host protocol

AaronFriel:
Enables building `pulumi-yaml` against trunk of `pulumi/pulumi`.


pkg/server/server.go
```diff
@@ -182,3 +183,16 @@ func (host *yamlLanguageHost) InstallDependencies(req *pulumirpc.InstallDependen
...
+	return &pulumirpc.GetProgramDependenciesResponse{}, nil
+}
+
+// About returns information about the runtime for this language.
+func (host *yamlLanguageHost) About(ctx context.Context, req *emptypb.Empty) (*pulumirpc.AboutResponse, error) {
+	return &pulumirpc.AboutResponse{
+		Executable: "pulumi-language-yaml",
+		Version:    version.Version,
+	}, nil
+}
```
AaronFriel, line 198: @Frassle this is what you'd expect, yeah?

aq17 (approved):

pkg/server/server.go
Frassle: Executable and version should be empty for yaml (you can just not set them, or set them to "")

AaronFriel: Blank? That's surprising to me, is this not conveyed back to the user?

Frassle: This is intended to report information about the executable we're using for the runtime itself, not information about the language plugin. So e.g. dotnet reports back something like "/usr/bin/dotnet", "6.0". 
yaml doesn't have an executable for the runtime, so it should just be blank. The logic in about handles the case that these can come back as `""`.

PR #296, opened 2022-07-26
Support setting default providers

### LINKED ISSUE #122: Copy resource provider's version/plugin download URL to resource when set
When the provider option is set on a resource, it may set the Version and PluginDownloadURL options. Those options should be copied to the resource to ensure that they both report the same plugin version.

Open question: Should we revert or change the functionality in #108 to not silently coalesce plugin versions? I think having two versions of the same plugin installed seems like an edge case, but it doesn't seem totally unreasonable that a customer's needs might require it.
See:
- https://github.com/pulumi/pulumi-yaml/pull/108

### LINKED ISSUE #293: Support provider default version setting along with `providers` project config
We'd like an example demonstrating that we can set the default provider version for invokes & resources within a program.

This repo uses AWS v4 in examples, we could create an example that sets the version to v5 and uses BucketV2. We could also find a deprecated/removed resource from another package and prove that we can acquire that and use it via default versioning.

For an EKS example, it would be interesting to set a default provider that uses the kubernetes config and uses that to create a namespace in the cluster.

Syntax should be similar to:

```yaml
resources:
  eks:
    type: eks:Cluster
  kubernetes:
    type: pulumi:pulumi:kubernetes
    defaultProvider: true
    properties:
      kubeconfig: ${eks.kubeconfig} # depends on EKS!
      enableServerSideApply: true
    options:
      version: 3.20
```

aq17:
Fixes #293 
Fixes #122

### DIFF
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -593,11 +593,6 @@ func (tc *typeCache) typeInvoke(ctx *evalContext, t *ast.InvokeExpr) bool {
 	if t.CallOpts.Provider != nil {
 		tc.typeExpr(ctx, t.CallOpts.Provider)
 	}
-	// } else if defaultPkgInfo, ok := ctx.defaultPackageInfo[pkg.Name()]; ok {
-	// 	expr, _ := ast.Interpolate("${" + defaultPkgInfo.providerResource.Value + "}")
-	// 	t.CallOpts.Provider = expr
-	// 	tc.typeExpr(ctx, t.CallOpts.Provider)
-	// }
 	if t.CallOpts.Version != nil {
 		tc.typeExpr(ctx, t.CallOpts.Version)
 	}
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -455,7 +455,7 @@ func (tc *typeCache) typeResource(r *runner, node resourceNode) bool {
 	}
 	if s := v.Syntax(); s != nil {
 		if o, ok := s.(*syntax.ObjectNode); ok {
-			validKeys := append(v.Fields(), "condition", "metadata", "defaultprovider")
+			validKeys := append(v.Fields(), "condition", "metadata")
 			fmtr := yamldiags.InvalidFieldBagFormatter{
 				ParentLabel: fmt.Sprintf("Resource %s", typ.String()),
 				MaxListed:   5,
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -455,7 +455,7 @@ func (tc *typeCache) typeResource(r *runner, node resourceNode) bool {
 	}
 	if s := v.Syntax(); s != nil {
 		if o, ok := s.(*syntax.ObjectNode); ok {
-			validKeys := append(v.Fields(), "condition", "metadata")
+			validKeys := append(v.Fields(), "condition", "metadata", "defaultprovider")
 			fmtr := yamldiags.InvalidFieldBagFormatter{
 				ParentLabel: fmt.Sprintf("Resource %s", typ.String()),
 				MaxListed:   5,
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -593,6 +593,11 @@ func (tc *typeCache) typeInvoke(ctx *evalContext, t *ast.InvokeExpr) bool {
 	if t.CallOpts.Provider != nil {
 		tc.typeExpr(ctx, t.CallOpts.Provider)
 	}
+	// } else if defaultPkgInfo, ok := ctx.defaultPackageInfo[pkg.Name()]; ok {
+	// 	expr, _ := ast.Interpolate("${" + defaultPkgInfo.providerResource.Value + "}")
+	// 	t.CallOpts.Provider = expr
+	// 	tc.typeExpr(ctx, t.CallOpts.Provider)
+	// }
 	if t.CallOpts.Version != nil {
 		tc.typeExpr(ctx, t.CallOpts.Version)
 	}
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -234,6 +234,33 @@ func setDefaultProviders(ctx *pulumi.Context, t *ast.TemplateDecl, r *runner) er
 			return true
 		},
 		VisitVariable: func(r *runner, node variableNode) bool {
+			v := node.Value
+			switch t := v.(type) {
+			case *ast.InvokeExpr:
+				pkgName := strings.Split(t.Token.Value, ":")[0]
+				if _, ok := defaultProviderInfoMap[pkgName]; !ok {
+					return true
+				}
+				defaultProviderInfo := defaultProviderInfoMap[pkgName]
+
+				if t.CallOpts.Version != nil &&
+					defaultProviderInfo.version.Value != t.CallOpts.Version.Value {
+					return false
+				}
+				if t.CallOpts.PluginDownloadURL != nil &&
+					defaultProviderInfo.pluginDownloadUrl.Value != t.CallOpts.PluginDownloadURL.Value {
+					return false
+				}
+
+				if t.CallOpts.Provider == nil {
+					expr, diags := ast.VariableSubstitution(defaultProviderInfo.providerResource.Value)
+					if diags.HasErrors() {
+						r.sdiags.diags = append(r.sdiags.diags, diags...)
+						return false
+					}
+					t.CallOpts.Provider = expr
+				}
+			}
 			return true
 		},
 		VisitConfig: func(r *runner, node configNode) bool {
@@ -1562,11 +1589,6 @@ func (ctx *evalContext) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}, b
 			ctx.error(t.Return, fmt.Sprintf("Unable to evaluate options Parent field: %+v", t.CallOpts.Parent))
 		}
 	}
-
-	// ctx.ctx.Log.Warn(fmt.Sprintf("pkgname is %s", pkgName), &pulumi.LogArgs{})
-	// dp, found := ctx.defaultPackageInfo[pkgName]
-	// ctx.ctx.Log.Warn(fmt.Sprintf("pkgname found? %v", found), &pulumi.LogArgs{})
-	// ctx.ctx.Log.Warn(fmt.Sprintf("pkg resource name %v", dp.providerResource), &pulumi.LogArgs{})
 	if t.CallOpts.Provider != nil {
 		providerOpt, ok := ctx.evaluateResourceValuedOption(t.CallOpts.Provider, "provider")
 		if ok {
@@ -1583,25 +1605,6 @@ func (ctx *evalContext) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}, b
 			ctx.error(t.Return, fmt.Sprintf("Unable to evaluate options Provider field: %+v", t.CallOpts.Provider))
 		}
 	}
-	// } else if defaultPkgInfo, ok := ctx.defaultPackageInfo[pkgName]; ok {
-	// 	// if no provider set, copy default provider to the resource
-	// 	expr, _ := ast.Interpolate("${" + defaultPkgInfo.providerResource.Value + "}")
-	// 	providerOpt, ok := ctx.evaluateResourceValuedOption(expr, "provider")
-	// 	if ok {
-	// 		if p, ok := providerOpt.(poisonMarker); ok {
-	// 			return p, true
-	// 		}
-	// 		provider := providerOpt.ProviderResource()
-	// 		if provider == nil {
-	// 			ctx.error(t.CallOpts.Provider, fmt.Sprintf("resource passed as Provider was not a provider resource '%s'", providerOpt))
-	// 		} else {
-	// 			opts = append(opts, pulumi.Provider(provider))
-	// 		}
-	// 	} else {
-	// 		ctx.error(t.Return, fmt.Sprintf("Unable to evaluate options Provider field: %+v", t.CallOpts.Provider))
-	// 	}
-	// }
-
 	performInvoke := ctx.lift(func(args ...interface{}) (interface{}, bool) {
 		// At this point, we've got a function to invoke and some parameters! Invoke away.
 		result := map[string]interface{}{}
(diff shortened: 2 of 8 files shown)


aq17:
This is still very much a work in progress – testing codegen is also dependent on resolving the strange failing tests for https://github.com/pulumi/pulumi/pull/10194/

Frassle:
This strikes me as not quite consistent.

Your allowing the declaring of provider resources, and setting them as 'default' but then resources aren't actually using the provider resource you've declared. They're just getting the version/url from that resource?

I think you should either:
A) Use "defaultProvider" to set the "provider" field on resources that don't have any other provider set. 
B) Have a new section to declare default providers which only allows setting of the version and url, nothing else.

aq17:
> This strikes me as not quite consistent.
> 
> Your allowing the declaring of provider resources, and setting them as 'default' but then resources aren't actually using the provider resource you've declared. They're just getting the version/url from that resource?
> 
> I think you should either: A) Use "defaultProvider" to set the "provider" field on resources that don't have any other provider set. B) Have a new section to declare default providers which only allows setting of the version and url, nothing else.

@Frassle I was trying to do that [here](https://github.com/pulumi/pulumi-yaml/blob/3e2d36c77e1769f36a9ec5b778855f82107562de/pkg/pulumiyaml/run.go#L913)

My understanding is there's two parts:
-If a default provider is declared, and the resource doesn't have the `provider` option set, we set it as the default provider — this is #293. If the resource instead sets the version/ url, warn if they match the default providers', error if not. If the resource doesn't have the version/ url, and the default provider specifies them, do we also need to copy them over, or is just setting the `provider` sufficient?
-if the resource does have the provider option set (regardless of whether it's default or not), we copy that provider's version/url if exists to the resource (and warn if the resource has also set the version/url as the same, error if not) — this is #122

iwahbe:
It feels like these are two different features. Is there a reason why they are in the same PR?

I image the two features as:
- https://github.com/pulumi/pulumi-yaml/issues/122: Add a static analysis pass, crawling all resources **and invokes** that have `options.provider` set, and assigning the resource the `version` and `pluginDownloadURL` of the provider. I would probably do this as a provider pass with a new walker after `TypeCheck` is called in `RunTemplate`. This should be relatively simple.
- https://github.com/pulumi/pulumi-yaml/issues/293: Add new syntax (simple). The rest will be a more complicated feature since it changes how the topological sort works. You will need to add an evaluator pass that hooks up default providers (for resources and invokes) that does not pre-suppose a topological sort, hook up the default providers, and then dump back into the normal pipeline (so the topological sort can be performed).

aq17:
> It feels like these are two different features. Is there a reason why they are in the same PR?
> 
> I image the two features as:
> 
> * [Copy resource provider's version/plugin download URL to resource when set #122](https://github.com/pulumi/pulumi-yaml/issues/122): Add a static analysis pass, crawling all resources **and invokes** that have `options.provider` set, and assigning the resource the `version` and `pluginDownloadURL` of the provider. I would probably do this as a provider pass with a new walker after `TypeCheck` is called in `RunTemplate`. This should be relatively simple.
> * [Support provider default version setting along with `providers` project config #293](https://github.com/pulumi/pulumi-yaml/issues/293): Add new syntax (simple). The rest will be a more complicated feature since it changes how the topological sort works. You will need to add an evaluator pass that hooks up default providers (for resources and invokes) that does not pre-suppose a topological sort, hook up the default providers, and then dump back into the normal pipeline (so the topological sort can be performed).

Spoke with @AaronFriel briefly and thought that we could preprocess in `RunTemplate` to store default+nondefault provider resource info (1version`,`pluginDownloadUrl`, and the actual resource state) within the runner. That way, we don't have to mess with the topological sort and can check each resource's options against the default or specified provider  in `registerResource`.

I _think_ the limitation of the static analysis pass is not being able to pull out the provider `version` or `url`, since `options.provider` is a `lateboundResource` that we cast to a `ProviderResourceState`

Frassle:
> https://github.com/pulumi/pulumi-yaml/issues/122: Add a static analysis pass, crawling all resources and invokes that have options.provider set, and assigning the resource the version and pluginDownloadURL of the provider. I would probably do this as a provider pass with a new walker after TypeCheck is called in RunTemplate. This should be relatively simple.

I don't think you want to assign the resource version and pluginDownloadURL here. Pretty sure the engine will just ignore them? Seems ok to have a diagnostic for the user though (you've set "provider" and "version", "version" will be ignored)

> https://github.com/pulumi/pulumi-yaml/issues/293: Add new syntax (simple). The rest will be a more complicated feature since it changes how the topological sort works. You will need to add an evaluator pass that hooks up default providers (for resources and invokes) that does not pre-suppose a topological sort, hook up the default providers, and then dump back into the normal pipeline (so the topological sort can be performed).

This feature changes depending on if your actually trying to get the engine to do "default providers" or if you just want yaml to assign an explicit provider by default.

iwahbe:
>> https://github.com/pulumi/pulumi-yaml/issues/122: Add a static analysis pass, crawling all resources and invokes that have options.provider set, and assigning the resource the version and pluginDownloadURL of the provider. I would probably do this as a provider pass with a new walker after TypeCheck is called in RunTemplate. This should be relatively simple.
>
>I don't think you want to assign the resource version and pluginDownloadURL here. Pretty sure the engine will just ignore them? Seems ok to have a diagnostic for the user though (you've set "provider" and "version", "version" will be ignored)

I'm not sure that the engine will ignore them. My understanding is that setting these options is the entire task of #122. Even if the engine ignores them, the type checker uses these properties to figure out which resource versions to check against. 

>> https://github.com/pulumi/pulumi-yaml/issues/293: Add new syntax (simple). The rest will be a more complicated feature since it changes how the topological sort works. You will need to add an evaluator pass that hooks up default providers (for resources and invokes) that does not pre-suppose a topological sort, hook up the default providers, and then dump back into the normal pipeline (so the topological sort can be performed).
>
>This feature changes depending on if your actually trying to get the engine to do "default providers" or if you just want yaml to assign an explicit provider by default.

I'm pretty sure we need to explicitly set providers, since Pulumi YAML is just the Go SDK under the hood, and I don't think there is a way to programmatically set default providers via the engine. I'll be very curios how this feature is supposed to work if we do it a different way.

iwahbe:
@aq17 For #122: 
The `Evaluator` operates on `*node`s, so you will have the fidelity necessary to do the static pass. You get `Expr` nodes for `Provider`, but you can cache those as you see them and the topological sorter will ensure you only hit nodes that you already cached.
https://github.com/pulumi/pulumi-yaml/blob/f3a244e33f5687e93593a5f3c7c3e99ae553dd28/pkg/pulumiyaml/run.go#L448-L453

For #293:
You will have to do a pre-processor scan for setting default providers, but it can't just be a linear scan of `Resources`, since you need to catch invokes as well. We need to know that a `aws` invoke must occur after the default `aws` provider (which might rely on other resources) is created. That means you must walk the AST to do the scan, make the adjustments, and then run the adjusted AST through the topological sorter.

Frassle:
> I'm not sure that the engine will ignore them. My understanding is that setting these options is the entire task of https://github.com/pulumi/pulumi-yaml/issues/122.

`getProviderReference` in pkg/resource/deploy/source_eval.go. If a provider urn is given that's the only thing used to decide the provider to use, if it's not given we build a provider request from package, version, and downloadURL.

aq17:
> > I'm not sure that the engine will ignore them. My understanding is that setting these options is the entire task of #122.
> 
> `getProviderReference` in pkg/resource/deploy/source_eval.go. If a provider urn is given that's the only thing used to decide the provider to use, if it's not given we build a provider request from package, version, and downloadURL.

If that's the case, #122 is essentially unnecessary once #293 is completed, except for the added diagnostics?

AaronFriel:
It does look like #122 might be unnecessary, but it does mean our SDK diverges from the way we codegen SDKs in other languages, which **always** set resource options for version (and pluginDownloadUrl if configured).

I think it's simpler to imitate the way codegen works in other languages, which ensures that if we change behavior in `pkg/resource/deploy/source_eval.go` at some point in the future, from the POV of the engine, the YAML language host looks just like a Go program using our Go SDK codegen.

AaronFriel:
Or in other words, this matrix indicates a risk of introducing subtle bugs down the road:

| Language | Always sets Version resource option |
|---|:---:|
| Go | ✅ |
| TS | ✅ |
| Python | ✅ |
| C# | ✅ |
| Java | ✅ |
| YAML | ❌ |

Frassle:
> It does look like https://github.com/pulumi/pulumi-yaml/issues/122 might be unnecessary, but it does mean our SDK diverges from the way we codegen SDKs in other languages, which always set resource options for version (and pluginDownloadUrl if configured).

Note that as far as I can see your plan for https://github.com/pulumi/pulumi-yaml/issues/122 is also divergent from what the other SDKs do?

Other SDKs have a hardcoded version that they always set _regardless_ of how the provider property is set. https://github.com/pulumi/pulumi-yaml/issues/122 looks to be saying "look at the provider property and copy it's version". Which firstly should be redundant, abut secondly isn't what the other SDKs do anyway, I'd probably lean to not doing that.

pkg/pulumiyaml/run.go
```diff
@@ -855,8 +885,27 @@ func (ctx *evalContext) registerResource(kvp resourceNode) (lateboundResource, b
 		} else {
 			overallOk = false
 		}
+	} else if defaultPkgInfo, ok := ctx.defaultPackageInfo[pkgName]; ok && !strings.HasPrefix(v.Type.Value, "pulumi:providers:") && false {
+		// if no provider set, copy default provider to the resource
+		expr, _ := ast.Interpolate("${" + defaultPkgInfo.providerResource.Value + "}")
+		providerOpt, ok := ctx.evaluateResourceValuedOption(expr, "provider")
```
aq17, line 891: This isn't quite working, as `providerOpt` resolves to a string 
    error: expected resource, got string
    error: error reading from server: EOF
 
```
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104a87b00]
    goroutine 83 [running]:
    github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/ast.(*StringExpr).Syntax(0x1400091c1a0?)
        <autogenerated>:1
    github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/ast.ExprError({0x104eb62d8?, 0x0?}, {0x14000462030, 0x30}, {0x0, 0x0})
        /Users/alexqiu/Desktop/pulumi-yaml/pkg/pulumiyaml/ast/expr.go:44 +0x4c
    github.com/pulumi/pulumi-yaml/pkg/pulumiyaml.(*evalContext).error(0x104ae0c83?, {0x104eb62d8?, 0x0?}, {0x14000462030?, 0x1?}) 
```

aq17:
@AaronFriel @iwahbe looks like `defaultProvider` works now (thanks for all the feedback 🙏🏼 ) – used this example
```
resources:
  myProvider:
    defaultProvider: true
    options:
      version: 5.9.0
    properties:
      region: us-west-2
    type: pulumi:providers:aws

  my-bucket:
    type: aws:s3:Bucket

variables:
  invokedRegion:
    Fn::Invoke:
      Function: aws:getRegion

outputs:
  actualRegion: ${myProvider.region}
  bucketRegion: ${my-bucket.region}
  invokedRegion: ${invokedRegion}
```
`pulumi convert` does not work, even after setting up my go workspace to use my local `pu/pu` and `pu/yaml` – from my understanding, this is expected because we'd need changes to codegen if we wanted this behavior of explicitly hooking up default providers?

Todo: add tests

pkg/pulumiyaml/run.go
```diff
@@ -176,16 +176,138 @@ func HasDiagnostics(err error) (syntax.Diagnostics, bool) {
...
+
+			if _, ok := defaultProviderInfoMap[pkgName]; !ok {
+				return true
+			}
+			defaultProviderInfo := defaultProviderInfoMap[pkgName]
+
+			if v.Options.Provider == nil {
+				if v.Options.Version != nil {
+					if defaultProviderInfo.version.Value != v.Options.Version.Value {
+						ctx.ctx.Log.Error("version set conflicts with default provider version", &pulumi.LogArgs{})
```
AaronFriel, line 218: I think instead of `ctx.ctx.Log.Error`, you should be able to call `ctx.errorf` in these situations, where the first argument is the expression where you detected an error.

I'd also simplify the logic here: if there is a default provider _period_, the user shouldn't be setting the version on resources. 

```go
    ctx.errorf(v.Options.Version, "Version conflicts with the default provider version. Try removing this option on resource %q.", resource name, ... etc)
```

AaronFriel: & a nit: it'd be good to be helpful and prescribe remedies in error messages, such as deleting the option.

AaronFriel (commented):
I think we can ship this without `convert` support for this milestone, and return to it by refactoring the runner so that we can run it in `load.go`. I've made some suggestions above to simplify the interface.

@iwahbe can you chat with @aq17 about how you'd want to integrate this rewriting pass into ejecting YAML? 

I think we'd want to eventually have a single function somewhere that takes a template, runs any rewriting transformations in a fixed order (so that it's uniform between `pulumi up` and `pulumi convert`), and then reports diagnostics.

The initial two passes would be this & typecheck, yeah? To make it really nice to use, we'd want to copy some convenience methods like `runner.errorf` as well.

pkg/pulumiyaml/run.go
```diff
@@ -176,16 +176,138 @@ func HasDiagnostics(err error) (syntax.Diagnostics, bool) {
 	}
 }
 
+func setDefaultProviders(ctx *pulumi.Context, t *ast.TemplateDecl, r *runner) error {
```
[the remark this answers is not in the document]
AaronFriel: @iwahbe flagging that right now evaluators/walkers are tied to the `runner` interface, and I think we'd want to refactor that out. I think we can do that next iteration?

pkg/pulumiyaml/analyser.go
```diff
@@ -455,7 +455,7 @@ func (tc *typeCache) typeResource(r *runner, node resourceNode) bool {
 	}
 	if s := v.Syntax(); s != nil {
 		if o, ok := s.(*syntax.ObjectNode); ok {
-			validKeys := append(v.Fields(), "condition", "metadata")
```
AaronFriel, line 458: I think this can be removed here, as `v.Fields()` should contain `defaultprovider`?

pkg/pulumiyaml/run.go
```diff
@@ -215,6 +321,12 @@ func (d *syncDiags) HasErrors() bool {
 	return d.diags.HasErrors()
 }
 
+type providerInfo struct {
+	version           *ast.StringExpr
+	pluginDownloadUrl *ast.StringExpr
```
AaronFriel, line 326: Try running `make lint` and check all the warnings raised.

pkg/pulumiyaml/run_options_test.go
```diff
@@ -106,6 +106,62 @@ resources:
 	}
 	err := pulumi.RunErr(func(ctx *pulumi.Context) error {
 		runner := newRunner(ctx, template, newMockPackageMap())
+		runner.setDefaultProviders()
```
AaronFriel, line 109: Use `requireNoErrors` or the appropriate `assert.` method to assert the return value isn't an error.

AaronFriel (approved):
LGTM once this passes the lint checks!

PR #305, opened 2022-08-15
Enable transformation & type checking passes in pulumi convert path

### LINKED ISSUE #302: Enable transformation & type checking passes in `pulumi convert` path
Following up on #296, we will likely want to refactor the Evaluator & runner interfaces so that we can easily invoke them from the `codegen` package.

aq17:
Fixes #302


aq17:
WIP – @iwahbe , instead of adding a new context without the runner, this moves the local state storage responsibilities into `evalContext` and removes the runner completely – is that reasonable?

pkg/pulumiyaml/run.go
```diff
@@ -196,11 +196,29 @@ func (r *runner) setDefaultProviders() error {
...
+	}
+
 	// Set roots
-	diags := r.Run(walker{
-		VisitResource: func(r *runner, node resourceNode) bool {
-			k, v := node.Key.Value, node.Value
-			ctx := r.newContext(node)
+	walker := walker{
+		VisitResource: func(ctx *evalContext, node resourceNode) bool {
+			_, v := node.Key.Value, node.Value
```
iwahbe, line 221:
```suggestion
			v := node.Value
```

iwahbe (commented):
>WIP – @iwahbe , instead of adding a new context without the runner, this moves the local state storage responsibilities into evalContext and removes the runner completely – is that reasonable?

This is completely reasonable. I think it's a good step. 

In the future, I would like the signature of `Run` to turn into something like the code below. That might be out of scope for this PR though. 
```go
func Run[T EvalContext](ctx T, e Evaluator[T]) syntax.Diagnostics

type EvalContext {
  error()
  newContext(node)
}
```

aq17:
> > WIP – @iwahbe , instead of adding a new context without the runner, this moves the local state storage responsibilities into evalContext and removes the runner completely – is that reasonable?
> 
> This is completely reasonable. I think it's a good step.
> 
> In the future, I would like the signature of `Run` to turn into something like the code below. That might be out of scope for this PR though.
> 
> ```go
> func Run[T EvalContext](ctx T, e Evaluator[T]) syntax.Diagnostics
> 
> type EvalContext {
>   error()
>   newContext(node)
> }
> ```

Once `pulumi-yaml` is bumped to >1.18 😃

pkg/pulumiyaml/analyser.go
```diff
@@ -964,10 +963,10 @@ func TypeCheck(r *runner) (Typing, syntax.Diagnostics) {
 }
 
 type walker struct {
-	VisitConfig   func(r *runner, node configNode) bool
-	VisitVariable func(r *runner, node variableNode) bool
-	VisitOutput   func(r *runner, node ast.PropertyMapEntry) bool
-	VisitResource func(r *runner, node resourceNode) bool
+	VisitConfig   func(ctx *evalContext, node configNode) bool
+	VisitVariable func(ctx *evalContext, node variableNode) bool
+	VisitOutput   func(node ast.PropertyMapEntry) bool
```
iwahbe, line 968: Even though we don't use it, we should probably pass a `*evalContext` to `VisitOutput` as well. It keeps the interface consistent. We also might want to use it later, or use it already in the lsp server.

iwahbe (commented):
I seem to remember that contexts were scoped to their nodes, and we used to figure out error origin. Does that machinery not exist anymore? CC @AaronFriel If this passes our diag tests, then I guess it does not.

Right now, it looks like we have just renamed `runner` to `evalContext`. I expected that contexts would create sub-contexts when they were passed through nodes. I think its fine to unify the types, but it doesn't really feel like a context anymore.

Your CHANGELOG entry implies that this will effect `pulumi convert`, but I don't see that. Is that part of this PR?

aq17:
> I seem to remember that contexts were scoped to their nodes, and we used to figure out error origin. Does that machinery not exist anymore? CC @AaronFriel If this passes our diag tests, then I guess it does not.
> 
> Right now, it looks like we have just renamed `runner` to `evalContext`. I expected that contexts would create sub-contexts when they were passed through nodes. I think its fine to unify the types, but it doesn't really feel like a context anymore.
> 
> Your CHANGELOG entry implies that this will effect `pulumi convert`, but I don't see that. Is that part of this PR?

If we want to limit the context's responsibilities to logging, we should decide where the runner's previous responsibilities should go?

AaronFriel:
I expect there should be something in between our current `evalContext` and `runner` implementation, that might inherit _some_ of the fields of a runner, but not all, that splits the difference and results in some sort of embedding hierarchy.

The specific thing we want to accomplish is separating out stateless transformations of the AST that can run without a `pulumi.Context` (what a `runner` has and `pulumi convert` won't have access to) and the actual *running* of the program.

```
// this is only used in run.go as it is the only place where we 
// have access to a pulumi.Context to construct one:
type runner struct {
	EvalContext // embed a root EvalContext
	
	ctx       *pulumi.Context
	
	// fields that aren't necessary for Evaluator/EvalContext. cwd, maybe?
}

// public type?
type EvalContext {
	Evaluator // embed evaluator which gives us diags, nodes, the sorted graph, etc.

	root   interface{}
}

// shared between multiple types
type Evaluator struct {
	// the remainder of the fields from runner
	sdiags syncDiags
}
```

Perhaps EvalContext & Evaluator can be merged into one type for simplicity.

iwahbe (commented):
I interpret the `runner`/`programEvaluator`/`evalContext` breakdown as follows:
- We have the `runner`, which stores the runtime state of a pulumi yaml program, as well as its contextual backing. I define runtime state as:
https://github.com/pulumi/pulumi-yaml/blob/7b764d047a588451c8cb58f4089d53c710dfec92/pkg/pulumiyaml/run.go#L344-L349
- We have the `evalContext`, which scopes a dive into part of a Pulumi YAML program to a specific root node.
- We have the `programEvaluator`, which drives runtime evaluation.

All three concepts have their own `diags` holder. Viewing the breakdown this way leads me to a couple of thoughts:

1. We only use runtime state (stored in `runner`) when actually running the program (driven by `programEvaluator`). Runtime state is bound to a specific `pulumi.Context`. Should we move our runtime state to be embedded in `programEvaluator` instead of `runner`?

2. All three structs have their own diagnostics. We should clarify which struct is responsible for which type of diagnostic. Do diagnostics of all `programEvaluator`, `evalContext` and the `runner` get printed? It makes sense to me to have `evalContext` handle diagnostics (or at least scoping them, since `evalContext` is used by all processes that walk the program). Can you write down the intended policy for diags (preferably in comments on each struct)?

3. It would be really helpful when reviewing if you added comments clarifying what each data type is conceptually for. Especially for WIP PRs like this one.

PR #307, opened 2022-08-16
CI check to ensure pu/pkg + pu/sdk versions are merged

### LINKED ISSUE #244: CI check to ensure that pulumi/pkg & /sdk versions are merged
YAML releases must reference a valid tag or commit that's part of the history of pulumi/pulumi to ensure that it is not deleted, causing an error we've seen internally with v0.3.0.

When users run `go mod tidy` against a package referencing YAML v0.3.0, they may see:

```
go: github.com/pulumi/pulumi-terraform-bridge/v3@v3.23.0 requires
        github.com/pulumi/pulumi-yaml@v0.3.0 requires
        github.com/pulumi/pulumi/pkg/v3@v3.31.1-0.20220429232336-81582c4dcb93: invalid version: unknown revision 81582c4dcb93
```

aq17:
Fixes #244


aq17:
@AaronFriel this runs `go mod tidy` which will fail if there are any errors (i.e. `unknown revision` due to invalid version), but if it succeeds, then just reverts the changes so it isn't a strict requirement to run `go mod tidy`

.github/workflows/stage-lint.yml
```diff
@@ -23,6 +23,13 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.46.2
+      - run: go mod tidy
+      - name: Undo go mod tidy changes if successful
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            git checkout ./go.mod && git checkout ./go.sum
```
AaronFriel, line 30: I think this would cause the check to exit with an error code and a helpful log message if `go.sum` isn't clean:

```suggestion
            echo "::error go.mod not tidy"
            exit 1
```

iwahbe (approved):
Looks good to me

Issue #306, opened 2022-08-16
Convert - Generated Go Program Contains Undefined `GetFloat` Method on Pulumi Configuration


phillipedwards:
### What happened?

Generated go program contains undefined Pulumi Configuration method `GetFloat`; Actual method is `GetFloat64`.

### Steps to reproduce

1. Create simple Pulumi yaml program with below definition.
```yaml
name: yaml-config-convert
runtime: yaml
description: A minimal Pulumi YAML program

configuration: 
  numberVal:
    type: Number
    default: 30
```
2. `pulumi convert --language go --out ./go`
3. See generated configuration access (`cfg.GetFloat`).
```golang
package main

import (
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		cfg := config.New(ctx, "")
		numberVal := 30
		if param := cfg.GetFloat("numberVal"); param != 0 {
			numberVal = param
		}
		return nil
	})
}
```

### Expected Behavior

Configuration access should utilize `GetFloat64` method.

### Actual Behavior

`GetFloat` method (undefined) is used.

### Additional context

Note: using a non-default value results in the same scenario for `RequireFloat` (should be `RequireFloat64`).

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

phillipedwards:
Aside from generating code which uses an undefined method, the generated method be `Get|Require(Int)` rather than `Get|Require(Float64)` based on the default value of `30`.

AaronFriel:
Interesting. There are two bugs here, codegen's misnamed type & Pulumi YAML's interpretation of the value.

@iwahbe It seems to me safe to generate int methods here. It seems like we could precisely choose the right config method to call by either performing type inference (could fail if the value is sent into both int & float properties) or by inserting cast/coercion calls where a resource expects a float.

iwahbe:
I think we should definitely add an `Int` type (https://github.com/pulumi/pulumi-yaml/pull/313) to the config, which would allow us to more precisely get config values.

[...]

PR #313, opened 2022-08-19
Introduce Int Type + Refactor Config Types

iwahbe:
This PR improves the config in 4 ways:
1. It refactors backing config types to use `schema.Type` instead of string equality. This allows us to share `DisplayType`, so that config error messages and type error messages display types the same way.
2. It introduces the `Int` type, providing a distinction between `Number`(float64) and `Int`(int) in the config type system.
3. It allows `Boolean` types to be correctly read from the config (since that was missing before).
4. Fixes a bug in displaying adhock object types.

Part of https://github.com/pulumi/pulumi-yaml/issues/306.

### DIFF
--- a/pkg/pulumiyaml/config/config.go
+++ b/pkg/pulumiyaml/config/config.go
@@ -5,6 +5,7 @@ package config
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
@@ -40,7 +41,6 @@ var (
 	BooleanList      = typ{&schema.ArrayType{ElementType: schema.NumberType}}
 	Int              = typ{schema.IntType}
 	IntList          = typ{&schema.ArrayType{ElementType: schema.IntType}}
-	Invalid          = typ{&schema.InvalidType{}}
 )
 
 type Types []Type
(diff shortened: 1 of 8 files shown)


pkg/pulumiyaml/run.go
```diff
@@ -809,6 +815,18 @@ func (ctx *evalContext) registerConfig(intm configNode) (interface{}, bool) {
...
+		if isSecretInConfig {
+			v, err = config.TrySecretInt(ctx.ctx, k)
+		} else {
+			v, err = config.TryInt(ctx.ctx, k)
+		}
+	case ctypes.Boolean:
+		if isSecretInConfig {
+			v, err = config.TrySecretBool(ctx.ctx, k)
+		} else {
+			v, err = config.TryInt(ctx.ctx, k)
```
aq17, line 828:
```suggestion
			v, err = config.TryBool(ctx.ctx, k)
```

aq17:
LGTM, supporting `int` -> `float` only seems like a good next step

pkg/pulumiyaml/run.go
iwahbe: Good catch!

pkg/pulumiyaml/config/config.go
```diff
@@ -120,3 +120,88 @@ func Parse(s string) (Type, bool) {
...
+				}
+				t2, err := TypeValue(v[i])
+				if err != nil {
+					return nil, err
+				}
+				return nil, &HeterogeneousListErr{t1, t2}
+			}
+		}
+		return expected, nil
+	case []float64:
```
aq17, line 210: Do the cases for `[]float64` and `[]int` need to be above the case for `[]interface{}`?

iwahbe: No. Go does not allow type coercion between `[]T` and `[]interface{}`.

pkg/pulumiyaml/config/config.go
```diff
@@ -120,3 +120,88 @@ func Parse(s string) (Type, bool) {
...
+	ErrUnexpectedType    = UnexpectedTypeErr{}
+)
+
+type HeterogeneousListErr struct {
+	T1 Type
+	T2 Type
+}
+
+func (e *HeterogeneousListErr) Error() string {
+	if e.T1 == nil || e.T2 == nil {
```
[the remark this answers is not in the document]
iwahbe: `nil` isn't a type, but we want an empty instance of a `HeterogeneousListErr` to be able to turn into a valid error message. Specifically, I'm thinking of the global variable defined here:
https://github.com/pulumi/pulumi-yaml/blob/ca907b2324a0a9b983fc8e11b389f531367a01d7/pkg/pulumiyaml/config/config.go#L125

aq17 (approved):
LGTM 👍🏼

PR #319, opened 2022-08-23
Fix pulumi convert panic on Fn::Split

### LINKED ISSUE #317: `pulumi convert` panics with certain YAML expressions
Given the following source file, which uses `Fn::Split` and `Fn::Select` to pull a portion of a string out of an output so it can be passed to another resource as an input:

```yaml
name: templates-static-website-azure-yaml
runtime: yaml
description: A static website on Azure

configuration:
  path:
    type: String
    default: ./site
  indexDocument:
    type: String
    default: index.html
  errorDocument:
    type: String
    default: error.html

# 👇
variables:
  originHostname:
    Fn::Select:
      - 2
      - Fn::Split:
        - /
        - ${account.primaryEndpoints.web}

resources:

  resource-group:
    type: azure-native:resources:ResourceGroup

  account:
    type: azure-native:storage:StorageAccount
    properties:
      resourceGroupName: ${resource-group.name}
      kind: StorageV2
      sku:
        name: Standard_LRS

  website:
    type: azure-native:storage:StorageAccountStaticWebsite
    properties:
      resourceGroupName: ${resource-group.name}
      accountName: ${account.name}
      indexDocument: ${indexDocument}
      error404Document: ${errorDocument}

  synced-folder:
    type: synced-folder:index:AzureBlobFolder
    properties:
      path: ${path}
      resourceGroupName: ${resource-group.name}
      storageAccountName: ${account.name}
      containerName: ${website.containerName}

  profile:
    type: azure-native:cdn:Profile
    properties:
      resourceGroupName: ${resource-group.name}
      sku:
        name: Standard_Microsoft

  endpoint:
    type: azure-native:cdn:Endpoint
    properties:
      resourceGroupName: ${resource-group.name}
      profileName: ${profile.name}
      isHttpAllowed: false
      isHttpsAllowed: true
      isCompressionEnabled: true
      contentTypesToCompress:
        - text/html
        - text/css
        - application/javascript
        - application/json
        - image/svg+xml
        - font/woff
        - font/woff2
      originHostHeader: ${originHostname}  # 👈
      origins:
        - name: ${account.name}
          hostName: ${originHostname}  # 👈

outputs:
  originURL: ${account.primaryEndpoints.web}
  cdnURL: https://${endpoint.hostName}
```

.. `pulumi convert` panics in the following way:

```bash
$ pulumi convert --out "$somelang" --language "$somelang"

================================================================================
The Pulumi CLI encountered a fatal error. This is a bug!
We would appreciate a report: https://github.com/pulumi/pulumi/issues/
Please provide all of the below text in your report.
================================================================================
Pulumi Version:   v3.38.0
Go Version:       go1.19
Go Compiler:      gc
Architecture:     amd64
Operating System: darwin
Panic:            fatal: A failure has occurred: unexpected builtin type *ast.SplitExpr
```

aq17:
Fixes #317

### DIFF
--- a/pkg/pulumiyaml/codegen/load.go
+++ b/pkg/pulumiyaml/codegen/load.go
@@ -200,6 +200,27 @@ func (imp *importer) importJoin(node *ast.JoinExpr) (model.Expression, syntax.Di
 	return call, diags
 }
 
+// importSplit imports a call to Fn::Split as a call to the `split` function.
+func (imp *importer) importSplit(node *ast.SplitExpr) (model.Expression, syntax.Diagnostics) {
+	var diags syntax.Diagnostics
+
+	delim, ddiags := imp.importExpr(node.Delimiter, nil)
+	diags.Extend(ddiags...)
+
+	source, sdiags := imp.importExpr(node.Source, nil)
+	diags.Extend(sdiags...)
+
+	if diags.HasErrors() {
+		return nil, diags
+	}
+
+	call := &model.FunctionCallExpression{
+		Name: "split",
+		Args: []model.Expression{delim, source},
+	}
+	return call, diags
+}
+
 // importFunctionCall imports a call to an AWS intrinsic function. The way the function is imported depends on the
 // function:
 //
@@ -296,6 +317,8 @@ func (imp *importer) importBuiltin(node ast.BuiltinExpr) (model.Expression, synt
 			Collection: values,
 			Key:        index,
 		}, diags
+	case *ast.SplitExpr:
+		return imp.importSplit(node)
 	case *ast.StackReferenceExpr:
 		stackName := node.StackName.Value
 		stackVar := imp.stackReferences[stackName]
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -28,3 +28,6 @@
 
 - Allow `bool` configuration type
   [#299](https://github.com/pulumi/pulumi-yaml/pull/299)
+
+- Fix `pulumi convert` panic on `Fn::Split`
+  [#319](https://github.com/pulumi/pulumi-yaml/pull/319)


pkg/pulumiyaml/codegen/load.go
```diff
@@ -296,6 +317,8 @@ func (imp *importer) importBuiltin(node ast.BuiltinExpr) (model.Expression, synt
 			Collection: values,
 			Key:        index,
 		}, diags
+	case *ast.SplitExpr:
```
t0yv0, line 320: Curious, was the panic originating in this default clause?

```
default:
		contract.Failf("unexpected builtin type %T", node)
		return nil, nil
```

I think `Failf` would cause a panic so it seems plausible.

We could perhaps consider changing that so that instead of panic-ing we return a placeholder expression like "TODO" and then a non-empty diagnostic. That'd fail more gracefully, WDYT?

Or are we close to covering every `case` here?

t0yv0:
Brilliant, very fast turn-around. Would it be possible to throw a test case in or we want to backlog that? I'm less familiar with this repo but elsewhere we have a mini-framework to work with test cases on programgen.

pkg/pulumiyaml/codegen/load.go
aq17: Yes – `Panic:            fatal: A failure has occurred: unexpected builtin type *ast.SplitExpr`

I think we should have every case covered here, it looks like `Fn::Split` may have been overlooked as it appears to be missing from our docs as well

aq17:
> Brilliant, very fast turn-around. Would it be possible to throw a test case in or we want to backlog that? I'm less familiar with this repo but elsewhere we have a mini-framework to work with test cases on programgen.

I think we could backlog it for now?

pkg/pulumiyaml/codegen/load.go
AaronFriel: It is. @aq17 could you ensure it's added to our docs? I think @cnunciato and a few other folks have noticed its absence.

AaronFriel (approved):

iwahbe (approved):
LGTM

PR #320, opened 2022-08-24
Enable transformation & type checking passes in pulumi convert path

### LINKED ISSUE #302: Enable transformation & type checking passes in `pulumi convert` path

aq17:
Fixes #302


pkg/pulumiyaml/run.go
```diff
@@ -276,21 +276,48 @@ func (r *runner) setDefaultProviders() error {
...
+	r := newRunner(t, loader)
 
-	err := r.setDefaultProviders()
+	cwd, err := os.Getwd()
 	if err != nil {
-		return err
+		r.sdiags.Extend(syntax.Error(nil, err.Error(), ""))
+		return nil, nil, err
+	}
+	r.variables[PulumiVarName] = map[string]interface{}{
```
aq17, line 289: @iwahbe I haven't moved the runtime state to `programEvaluator` for now, so we can easily initialize a `programEvaluator` with just a ctx (line 639)

pkg/pulumiyaml/codegen/convert.go
```diff
@@ -49,7 +49,17 @@ func ConvertTemplateIL(template *ast.TemplateDecl, loader schema.ReferenceLoader
 		loader = schema.NewPluginLoader(host)
 	}
 
-	templateBody, tdiags := ImportTemplate(template, pulumiyaml.NewPackageLoaderFromSchemaLoader(loader))
+	pkgLoader := pulumiyaml.NewPackageLoaderFromSchemaLoader(loader)
+	_, tdiags, err := pulumiyaml.PrepareTemplate(template, pkgLoader)
```
AaronFriel, line 53: Might help to comment why it's OK to discard the runner return value here - I think it's because (a.) we aren't going to execute the template and (b.) the loader stored on the runner is already in our local scope and reused?

AaronFriel (approved):
This looks great. Only a nit on the discarded return value before.

PR #324, opened 2022-08-26
fix: Type checker inaccurately allows using 'id' property of a component resource

AaronFriel:
Fixes the invalid Pulumi YAML program execution in #323, with this change the program in that issue will report an error like so:

```
pulumi:pulumi:Stack (k8s-aws-yaml-dev):
    error: id does not exist on vpc
    
      on Pulumi.yaml line 37:
      37:       vpcId: ${vpc.id}
    
    Existing properties are: eips, urn, vpc, vpcId, routes and 9 others
```

Hopefully pointing them in the direction of using `vpc.vpcId`.


iwahbe (approved):
LGTM

PR #331, opened 2022-09-09
Set highest package version on all resources

aq17:
This sets on all resources the highest version when converting from YAML to PCL on `pulumi convert`, if at least one resource has a version set

Necessary for changes here https://github.com/pulumi/pulumi/pull/10194/

### DIFF
--- a/pkg/pulumiyaml/codegen/load.go
+++ b/pkg/pulumiyaml/codegen/load.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"sort"
 
+	"github.com/blang/semver"
 	"github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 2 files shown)


pkg/pulumiyaml/codegen/load.go
```diff
@@ -602,8 +603,34 @@ func (imp *importer) importVariable(kvp ast.VariablesMapEntry) (model.BodyItem,
 	}, diags
 }
 
+// gets the highest version specified for a package
+// TODO: support invokes
```
AaronFriel, line 607: I'd also follow up with this and make sure we track the pluginDownloadUrl

AaronFriel (approved):
LGTM - I'm OK with shipping these incrementally to get version working first, then pluginDownloadUrl, then invokes.

PR #332, opened 2022-09-09
Set highest pkg version+pluginDownloadURL on `pulumi convert`

aq17:
Extends https://github.com/pulumi/pulumi-yaml/pull/331 by setting `pluginDownloadURL` as well on resources

Todo: invokes

### DIFF
--- a/pkg/pulumiyaml/codegen/load.go
+++ b/pkg/pulumiyaml/codegen/load.go
@@ -33,6 +33,11 @@ type importer struct {
 	outputs         map[string]*model.Variable
 }
 
+type packageInfo struct {
+	version           string
+	pluginDownloadURL string
+}
+
 // null represents PCL's builtin `null` variable
 var null = &model.Variable{
 	Name:         "null",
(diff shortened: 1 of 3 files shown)


pkg/pulumiyaml/codegen/load.go
```diff
@@ -712,10 +725,14 @@ func (imp *importer) importResource(kvp ast.ResourcesMapEntry, highestPkgVersion
...
-	if resource.Options.Version != nil {
+	if p, ok := highestPkgInfo[pkg.Name()]; ok {
 		resourceOptions.Body.Items = append(resourceOptions.Body.Items, &model.Attribute{
 			Name:  "version",
-			Value: quotedLit(highestPkgVersion[pkg.Name()]),
+			Value: quotedLit(p.version),
+		})
+		resourceOptions.Body.Items = append(resourceOptions.Body.Items, &model.Attribute{
+			Name:  "pluginDownloadURL",
+			Value: quotedLit(p.pluginDownloadURL),
```
iwahbe, line 735: We should probably check that `p.pluginDownloadURL != ""`. Same for `p.version`

pkg/tests/transpiled_examples/webserver-json/program.pp
```diff
@@ -1,51 +1,57 @@
...
+        version = "4.37.1"
+        pluginDownloadURL = ""
+    }
 }
 
 output instanceId {
-	__logicalName = "InstanceId"
-	value = webServer.id
+    __logicalName = "InstanceId"
+    value = webServer.id
```
iwahbe, line 46: Why did this happen?

iwahbe (approved):
The change looks good, pending 🟢 tests.

AaronFriel (commented):
Could you update our examples with a test that uses a pluginDownloadUrl? You might need to ask in Slack for a good example of a third party provider we would want to test.

aq17:
> Could you update our examples with a test that uses a pluginDownloadUrl? You might need to ask in Slack for a good example of a third party provider we would want to test.

Sounds good!

PR #339, opened 2022-09-20
Fix generated Pulumi.yaml on convert

aq17:
Fixes `pu/pu`#[10795](https://github.com/pulumi/pulumi/issues/10795)

### DIFF
--- a/pkg/pulumiyaml/codegen/eject.go
+++ b/pkg/pulumiyaml/codegen/eject.go
@@ -17,6 +17,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
 
+var ProjectKeysToOmit = []string{"configuration", "resources", "outputs"}
+
 // Eject on a YAML program directory returns a Pulumi Project and a YAML program which has been
 // parsed and converted to the intermediate PCL language
 //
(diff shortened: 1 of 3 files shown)


Frassle (commented):
Change looks ok, is there a framework in pulumi-yaml for testing eject by itself yet?

aq17:
> Change looks ok, is there a framework in pulumi-yaml for testing eject by itself yet?

Not standalone for eject, but could possibly add a check [here](https://github.com/pulumi/pulumi-yaml/blob/main/pkg/tests/example_transpile_test.go#L283-L285)?

pkg/pulumiyaml/codegen/eject.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
 
+var ProjectKeysToOmit = []string{"configuration", "resources", "outputs"}
```
AaronFriel, line 20: I think `variables` would also be excluded here?

AaronFriel (commented):
I believe there's another key to omit, otherwise a fairly clear change here.

AaronFriel (approved):
bors merge

AaronFriel (approved):
LGTM!

PR #340, opened 2022-09-21
Typecheck nested properties

### LINKED ISSUE #330: Typechecker allows non-existent fields in embedded objects
### What happened?

See https://github.com/pulumi/pulumi/issues/10466 for details.

### Steps to reproduce

Run `pulumi up` on 
```yaml
name: yaml-test
runtime: yaml
resources:
  patch1:
    type: kubernetes:apps/v1:DeploymentPatch
    properties:
      metadata:
        name: "coredns"
        namespace: "kube-system"
        annotations:
          pulumi.com/patchForce: "true"
      spec:
        template:
          containers:
            name: "coredns"
            # Use the version retrieved earlier as the image tag here!
            image: "602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/coredns:v1.8.7-eksbuild.1"
```
With `pulumi-eks = v0.41.2`

### Expected Behavior

We should reject an invalid program.

### Actual Behavior

We let it in.

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

aq17:
Fixes #330 

todo: add test

### DIFF
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -375,7 +375,7 @@ func (tc *typeCache) typeResource(r *Runner, node resourceNode) bool {
 		FieldsAreProperties: true,
 	}
 
-	tc.typePropertyEntries(ctx, k, fmtr, pkg, v.Properties.Entries, hint.Resource.InputProperties)
+	tc.typePropertyEntries(ctx, k, fmtr, v.Properties.Entries, hint.Resource.InputProperties)
 
 	tc.registerResource(k, node.Value, hint)
 
@@ -402,7 +402,7 @@ func (tc *typeCache) typeResource(r *Runner, node resourceNode) bool {
 		MaxElements:         5,
 		FieldsAreProperties: true,
 	}
-	tc.typePropertyEntries(ctx, k, fmtr, pkg, v.Get.State.Entries, hint.Resource.Properties)
+	tc.typePropertyEntries(ctx, k, fmtr, v.Get.State.Entries, hint.Resource.Properties)
 
 	// Check for extra fields that didn't make it into the resource or resource options object
 	options := ResourceOptionsTypeHint()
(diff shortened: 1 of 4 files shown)


pkg/pulumiyaml/analyser.go
```diff
@@ -490,6 +492,29 @@ func (tc *typeCache) typePropertyEntries(ctx *evalContext, resourceName string,
 			subject := kvp.Key.Syntax().Syntax().Range()
 			ctx.addErrDiag(subject, summary, detail)
 		} else {
+			if obj, ok := kvp.Value.(*ast.ObjectExpr); ok {
+				prop := propMap[kvp.Key.Value]
```
iwahbe, line 496: This variable already exists as `typ`. It's bound during the if check.

iwahbe (changes requested):
I don't think that `typePropertyEntries` is the right place to put this. It serves to aggregate properties into a `schema.Type` and hand that off to `assertTypeAssignable`. I think the fix needs to live in `assertTypeAssignable`.

I'm pretty sure there is a bug in 

https://github.com/pulumi/pulumi-yaml/blob/60bfab81632daf5cb945e04b57412dc135fed4d8/pkg/pulumiyaml/analyser.go#L295-L321

`assertTypeAssignable` handles complications like `schema.Union`, which we need to account for.

aq17:
> I don't think that `typePropertyEntries` is the right place to put this. It serves to aggregate properties into a `schema.Type` and hand that off to `assertTypeAssignable`. I think the fix needs to live in `isAssignable`.
> 
> I'm pretty sure there is a bug in
> 
> https://github.com/pulumi/pulumi-yaml/blob/60bfab81632daf5cb945e04b57412dc135fed4d8/pkg/pulumiyaml/analyser.go#L295-L321
> 
> `isAssignable` handles complications like `schema.Union`, which we need to account for.

I'm unsure how you get the subproperties (`from`) within just `isAssignable` though. If the`from` property is an object, we can get its subproperties and recursively call `isAssignable` (where `from` is now `to`), but what do we compare it to?

iwahbe:
`to` would also be an object, `isAssignable` just asserts that an object fits within another object.

PR #342, opened 2022-09-22
Allow interpolations for `AssetOrArchive` function values

### LINKED ISSUE #341: `Fn::StringAsset` doesn't handle expressions
### What happened?

In other languages, I can declare a `StringAsset` using, say, an output from another resource:

```typescript
const app = new azure_native.web.WebApp("app", {
    // ...
});

const configFile = new azure_native.storage.Blob("config.json", {
    source: app.defaultHostName
        .apply(hostname => new pulumi.asset.StringAsset(JSON.stringify({ 
            apiEndpoint: `https://${hostname}/api` 
        }))),
    contentType: "application/json",
    // ...
});
```

When I try this with YAML, though:

```yaml
configFile:
    Fn::StringAsset: '{ "apiEndpoint": https://${app.defaultHostName}/api }'
```

I get an error:

```
error: could not load yaml program: failed to load the template: Pulumi.yaml:34,22-75: 
    The argument to Fn::StringAsset must be a string literal; 
```

Ideally, YAML would support interpolation like this as well.

### Steps to reproduce

See above.

### Expected Behavior

The `StringAsset` is created in the usual way.

### Actual Behavior

See above.

### Output of `pulumi about`

```
CLI
Version      3.40.1
Go Version   go1.19.1
Go Compiler  gc

Host
OS       darwin
Version  12.5.1
Arch     x86_64
```

### Additional context

_No response_

### Contributing

aq17:
Fixes #341

### DIFF
--- a/pkg/pulumiyaml/ast/expr.go
+++ b/pkg/pulumiyaml/ast/expr.go
@@ -352,12 +352,12 @@ func ParseExpr(node syntax.Node) (Expr, syntax.Diagnostics) {
 	}
 }
 
-var assetOrArchiveFunctions = map[string]func(node syntax.Node, k, v *StringExpr) Expr{
-	"Fn::StringAsset":   func(node syntax.Node, k, v *StringExpr) Expr { return StringAssetSyntax(node, k, v) },
-	"Fn::FileAsset":     func(node syntax.Node, k, v *StringExpr) Expr { return FileAssetSyntax(node, k, v) },
-	"Fn::RemoteAsset":   func(node syntax.Node, k, v *StringExpr) Expr { return RemoteAssetSyntax(node, k, v) },
-	"Fn::FileArchive":   func(node syntax.Node, k, v *StringExpr) Expr { return FileArchiveSyntax(node, k, v) },
-	"Fn::RemoteArchive": func(node syntax.Node, k, v *StringExpr) Expr { return RemoteArchiveSyntax(node, k, v) },
+var assetOrArchiveFunctions = map[string]func(node syntax.Node, k *StringExpr, v Expr) Expr{
+	"Fn::StringAsset":   func(node syntax.Node, k *StringExpr, v Expr) Expr { return StringAssetSyntax(node, k, v) },
+	"Fn::FileAsset":     func(node syntax.Node, k *StringExpr, v Expr) Expr { return FileAssetSyntax(node, k, v) },
+	"Fn::RemoteAsset":   func(node syntax.Node, k *StringExpr, v Expr) Expr { return RemoteAssetSyntax(node, k, v) },
+	"Fn::FileArchive":   func(node syntax.Node, k *StringExpr, v Expr) Expr { return FileArchiveSyntax(node, k, v) },
+	"Fn::RemoteArchive": func(node syntax.Node, k *StringExpr, v Expr) Expr { return RemoteArchiveSyntax(node, k, v) },
 }
 
 // Attempts to parse an asset or archive. These are not normal `Fn::*` objects
@@ -373,13 +373,7 @@ func tryParseAssetOrArchive(k, v Expr) (Expr, syntax.Diagnostics, bool) {
 	}
 
 	if fn, ok := assetOrArchiveFunctions[fnName.Value]; ok {
-		s, ok := v.(*StringExpr)
-		if !ok {
-			diags.Extend(syntax.NodeError(v.Syntax(), fmt.Sprintf("The argument to %s must be a string literal", fnName.Value), ""))
-			return nil, diags, true
-		}
-		return fn(k.Syntax(), fnName, s), diags, true
-
+		return fn(k.Syntax(), fnName, v), diags, true
 	}
 
 	// Not a asset or archive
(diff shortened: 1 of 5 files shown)


pkg/pulumiyaml/ast/expr.go
```diff
@@ -375,8 +375,12 @@ func tryParseAssetOrArchive(k, v Expr) (Expr, syntax.Diagnostics, bool) {
...
 		s, ok := v.(*StringExpr)
 		if !ok {
-			diags.Extend(syntax.NodeError(v.Syntax(), fmt.Sprintf("The argument to %s must be a string literal", fnName.Value), ""))
-			return nil, diags, true
+			i, ok := v.(*InterpolateExpr)
+			if !ok {
+				diags.Extend(syntax.NodeError(v.Syntax(), fmt.Sprintf("The argument to %s must be a string literal or interpolated expression", fnName.Value), ""))
+				return nil, diags, true
+			}
+			return fn(k.Syntax(), fnName, String(i.Syntax().String())), diags, true
```
AaronFriel, line 383: I think this will result in the string containing the (un-interpolated) `"${...}`. 

Can you add a unit test like the one here https://github.com/pulumi/pulumi-yaml/blob/983b9a7894b7c18ed7c2cc49665c1cd55e2f5a6f/pkg/pulumiyaml/run_test.go#L354

AaronFriel (commented):
I have a hunch this won't quite do the right thing because of the string casting & conversion, and that the node constructors (lines 355-360 in expr.go) and evaluator for these functions will need to be updated.

cnunciato:
Checking in on this one. Is it still being worked on?

aq17:
> Checking in on this one. Is it still being worked on? (It's blocking an architecture template.)

@cnunciato Yes, apologies for the delay - just need to resolve the failing test

cnunciato:
Thank you! Super appreciate it @aq17!

pkg/pulumiyaml/run.go
```diff
@@ -1861,6 +1864,101 @@ func (e *programEvaluator) evaluateBuiltinSecret(s *ast.SecretExpr) (interface{}
...
+			}
+			return pulumi.NewRemoteAsset(value), true
+
+		}
+		return e.error(s, "unhandled expression")
+	})
+
+	return createAssetArchiveF(v)
+}
+
```
iwahbe, line 1912:
```suggestion
// sanitizePath checks that `path` exists and points to either a file within `./` or is constant and absolute. This provides security when running Pulumi YAML programs.
```

pkg/pulumiyaml/run.go
```diff
@@ -1861,6 +1864,101 @@ func (e *programEvaluator) evaluateBuiltinSecret(s *ast.SecretExpr) (interface{}
...
+	_, isConstant := s.(*ast.StringExpr)
+	v, b := e.evaluateExpr(s)
+	if !b {
+		return nil, false
+	}
+
+	createAssetArchiveF := e.lift(func(args ...interface{}) (interface{}, bool) {
+		value, ok := args[0].(string)
+		if !ok {
+			return e.error(s, fmt.Sprintf("Argument to Fn::ReadFile must be a string, got %v", reflect.TypeOf(args[0])))
```
iwahbe, line 1877: This should say `Fn::*`, not just `Fn::ReadFile`.

pkg/pulumiyaml/run.go
```diff
@@ -1861,6 +1864,101 @@ func (e *programEvaluator) evaluateBuiltinSecret(s *ast.SecretExpr) (interface{}
...
+			return pulumi.NewRemoteAsset(value), true
+
+		}
+		return e.error(s, "unhandled expression")
+	})
+
+	return createAssetArchiveF(v)
+}
+
+func (e *programEvaluator) sanitizePath(value string, isConstant bool) (string, error) {
```
iwahbe, line 1913: I think `path` would be a much cleaner name then `value`.
```suggestion
func (e *programEvaluator) sanitizePath(path string, isConstant bool) (string, error) {
```

iwahbe (approved):
Changes look good. We can merge once tests go 🟢.

I'm wondering if it would help to introduce an interface like:
```go
type AssetOrArchive interface {
  New(value string, sanatizer func(string, bool) (string, error)) (any, error)
  Name() string
  isAssetOrArchive()
}
```

It would avoid multiple sets of matching `switch` statements.

iwahbe:
> > Changes look good. We can merge once tests go 🟢.
> > I'm wondering if it would help to introduce an interface like:
> > ```go
> > type AssetOrArchive interface {
> >   New(value string, sanatizer func(string, bool) (string, error)) (any, error)
> >   Name() string
> >   isAssetOrArchive()
> > }
> > ```
> > 
> > It would avoid multiple sets of matching `switch` statements.
> 
> @iwahbe could we just add `New(value string, sanatizer func(string, bool) (string, error)) (any, error)` on [AssetOrArchiveExpr](https://github.com/pulumi/pulumi-yaml/blob/main/pkg/pulumiyaml/ast/expr.go#L582)

Maybe. I think we want to be able to reference runtime info in these methods (for debugging if nothing else), and thus would prefer this interface to live in the `pulumiyaml` module, while `AssetOrArchiveExpr` is in `ast`. I don't think that `ast` can reach back into into `pulumiyaml` to see runtime representation types. I'd have to play with it to be sure.

AaronFriel (approved):

PR #360, opened 2022-10-04
Add integration test to check type error display

### LINKED ISSUE #347: Add integration tests that fail type checking
## Hello!

## Issue details

We had a panic on all programs that didn't type check (#345), and this merged into master. We should catch such problems in automated tests.

iwahbe:
This test is unfortunately and necessarily fragile, since it tests what the user is displayed instead of a structured result.

The test will not pass before https://github.com/pulumi/pulumi/pull/10918 merges.

Fixes #347


iwahbe:
https://github.com/pulumi/pulumi/pull/10918 has merged, so we can merge this PR after the next release (which should be tomorrow).

pkg/tests/integration_test.go
```diff
@@ -0,0 +1,65 @@
...
+Cannot assign '{length: string, lower: number}' to 'random:index/randomString:RandomString':
+
+  length: Cannot assign type 'string' to type 'integer'
+
+  lower: Cannot assign type 'number' to type 'boolean'
+
+Resources:
+    + 1 to create
+
+`, stackName), stderr)
```
dixler, line 61: on first glance, I'd suggest asserting that

> length: Cannot assign type 'string' to type 'integer'

and

>   lower: Cannot assign type 'number' to type 'boolean'

are in the output since this seems fragile if we change the output and those errors are controlled by this repository.

iwahbe: We can do that.

pkg/tests/integration_test.go
```diff
@@ -0,0 +1,30 @@
...
+//nolint:paralleltest // uses parallel programtest
+func TestTypeCheckError(t *testing.T) {
+	testWrapper(t, integrationDir("type-fail"), ExpectFailure, StderrValidator{
+		f: func(t *testing.T, stderr string) {
+			assert.Contains(t, stderr,
+				`Cannot assign '{length: string, lower: number}' to 'random:index/randomString:RandomString':
+
+  length: Cannot assign type 'string' to type 'integer'
+
+  lower: Cannot assign type 'number' to type 'boolean'
```
dixler, line 25: nit. I'd break these out into 3 different `assert.Contains` since I'm not sure if the ordering is guaranteed or if we want the test to guarantee the ordering.

dixler (approved):
LGTM other than a nit around possible future test flakiness, but I trust your judgement.

pkg/tests/integration_test.go
iwahbe: The ordering is guaranteed and important. I'm happy to have this test scream at us if the ordering changes.

Remember, the invalid part of the test file looks like this:
```yaml
  randomString:
     type: random:RandomString
     properties:
       length: "three"
       lower: 7
```
We want that `properties` (`{length: string, lower: number}`) is invalid first, and then, indented and in order, the reasons why the main property assignment is invalid.

iwahbe:
Blocked on https://github.com/pulumi/pulumi/pull/10965.

AaronFriel (approved):
LGTM

PR #362, opened 2022-10-08
Warn on non-camelCase naming

### LINKED ISSUE #355: Make case sensitivity consistent
## Hello!

### The Problem
Pulumi YAML is sometimes case sensitive, sometimes not. When it is case sensitive, different capitalization schemes are required. We need to decide on a consistent capitalization scheme. Any decision is better than what we have now.

Our decision is complicated by the fact that Pulumi YAML consumes certain fields, but passes through others. Consumed fields can be cross–checked against the relevant schema to correct for capitalization. Fields that are passed through must be case sensitive.
This short program gives an example of the problem:
Screen Shot 2022-10-07 at 2 36 37 PM

This is not intuitive right now, but it definitely should be.
### The Solution
TLDR: The least confusing option is to make everything case sensitive.

We must make certain areas case sensitive, such as string values and the keys to maps. The values are entirely pass through, and there is no way to correct invalid casing. Because of the uniform syntax of YAML, it's not always clear when you are in a case sensitive setting. We can eliminate this question by teaching users that the answer is always: yes, it is case sensitive. This will improve the writability of Pulumi YAML.

We currently require both CamelCase and camelCase styles when interacting with functions and properties respectively. Other areas, such as Pulumi YAML top level keys, give the user the ability to choose which style they prefer. We have learned recently that it is beneficial for the language to enforce a preferred naming style (Rust, Go) because it forces similar programs to look similar. We should adopt this best practice to improve the readability of Pulumi YAML.

Under this solution, we would standardize all Pulumi YAML keys and values on camelCase, which matches with the property names in our schemas.

This would be a breaking change for our existing ecosystem. We would roll out a release that generates warnings for invalid casing, and then another release that would turn these warnings into errors.

### Affected area/feature

The `ast` and `syntax` modules.

iwahbe:
Fixes #355 

Generated warnings look like this:

```yaml
name: yaml
runtime: yAML
Resources:
  provider:
    type: pulumi:providers:aws
    PrOperties:
      region: us-east-1
    DefaultProvider: true
  bucket:
    Type: aws:s3:BucketV2
```

```
Diagnostics:
  pulumi:pulumi:Stack (yaml-dev):
    Warning: 'Resources' looks like a miscapitalization of 'resources'
      on Pulumi.yaml line 3:
       3: Resources:
    Pulumi YAML will enforce camelCase capitalization by GA. See https://github.com/pulumi/pulumi-yaml/issues/355 for details.
    Warning: 'PrOperties' looks like a miscapitalization of 'properties'
      on Pulumi.yaml line 6:
       6:     PrOperties:
    Pulumi YAML will enforce camelCase capitalization by GA. See https://github.com/pulumi/pulumi-yaml/issues/355 for details.
    Warning: 'DefaultProvider' looks like a miscapitalization of 'defaultProvider'
      on Pulumi.yaml line 8:
       8:     DefaultProvider: true
    Pulumi YAML will enforce camelCase capitalization by GA. See https://github.com/pulumi/pulumi-yaml/issues/355 for details.
    Warning: 'Type' looks like a miscapitalization of 'type'
      on Pulumi.yaml line 10:
      10:     Type: aws:s3:BucketV2
    Pulumi YAML will enforce camelCase capitalization by GA. See https://github.com/pulumi/pulumi-yaml/issues/355 for details.
```

### DIFF
--- a/pkg/pulumiyaml/syntax/diags.go
+++ b/pkg/pulumiyaml/syntax/diags.go
@@ -135,6 +135,6 @@ func UnexpectedCasing(rng *hcl.Range, expected, found string) *Diagnostic {
 		return nil
 	}
 	summary := fmt.Sprintf("'%s' looks like a miscapitalization of '%s'", found, expected)
-	detail := "Pulumi YAML will enforce camelCase capitalization by GA. See https://github.com/pulumi/pulumi-yaml/issues/355 for details."
+	detail := "A future version of Pulumi YAML will enforce camelCase fields. See https://github.com/pulumi/pulumi-yaml/issues/355 for details."
 	return Warning(rng, summary, detail)
 }
--- a/pkg/pulumiyaml/syntax/diags.go
+++ b/pkg/pulumiyaml/syntax/diags.go
@@ -97,7 +97,11 @@ func (d Diagnostics) Error() string {
 // Extend appends the given list of diagnostics to the list.
 func (d *Diagnostics) Extend(diags ...*Diagnostic) {
 	if len(diags) != 0 {
-		*d = append(*d, diags...)
+		for _, diag := range diags {
+			if diag != nil {
+				*d = append(*d, diag)
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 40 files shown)


pkg/pulumiyaml/syntax/diags.go
```diff
@@ -121,3 +125,16 @@ func (d Diagnostics) Unshown() *Diagnostics {
...
+// Inform the user that they did not conform with the expected capitalization style. If
+// `expected` matches `found`, then `nil` is returned. This allows
+// `Diagnostics.Extend(UnexpectedCasing(location, expected, found))` without checking if
+// expected equals found.
+func UnexpectedCasing(rng *hcl.Range, expected, found string) *Diagnostic {
+	if expected == found {
+		return nil
+	}
+	summary := fmt.Sprintf("'%s' looks like a miscapitalization of '%s'", found, expected)
+	detail := "Pulumi YAML will enforce camelCase capitalization by GA. See https://github.com/pulumi/pulumi-yaml/issues/355 for details."
```
AaronFriel, line 138: We may punt on enforcing this by GA, to allow more time for users to fix programs.

```suggestion
	detail := "A future version of Pulumi YAML will enforce camelCase fields. See https://github.com/pulumi/pulumi-yaml/issues/355 for details."
```

AaronFriel (approved):
LGTM, left a nit on the warning message.

Issue #251, opened 2022-06-16
Migrate backend of code generation tests to pulumi/pulumi


AaronFriel:
We use an intermediate representation in Pulumi YAML to pass information to `pulumi convert`, and then the code generation to target languages occurs in code owned by the pulumi/pulumi repo.

PR #365, opened 2022-10-12
Sync PCL files with pu/pu on release

aq17:
Keep PCL test files in both repos in sync after they are decoupled per #251

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,18 @@
+name: Sync Files
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  convert-test-sync:
+    name: Sync PCL test files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Run GitHub File Sync
+        uses: BetaHuhn/repo-file-sync-action@40da4d517e0a744fce9e8aed03e63d6c8bafbdc0 # v1.17.21
+        with:
+          GH_PAT: ${{ secrets.PULUMI_BOT_TOKEN }}
+          TEAM_REVIEWERS: pulumi/platform
--- a/.github/sync.yml
+++ b/.github/sync.yml
@@ -1,0 +1,3 @@
+pulumi/pulumi:
+  - source: ../pkg/tests/transpiled_examples
+    dest: pkg/codegen/testing/test/testdata/transpiled_examples


aq17:
@AaronFriel could this be a better approach to keeping the PCL files in sync? It would also help keep `pulumi/pulumi` up to date with the latest `pulumi-yaml` release as well

iwahbe (approved):
The way I'm reading this, on every release we sync PCL files from our pulumi-yaml transpiled examples to pulumi/pulumi's transpired-examples. If there are diffs, we open a PR against pulumi/pulumi and tag platform for review.

Assuming that the ☝️ is correct, this looks good.

P.S. We all got an email that pulumi/platform was deleted. Was that resolved? If not this will tag a non-existent team for review.

aq17:
> The way I'm reading this, on every release we sync PCL files from our pulumi-yaml transpiled examples to pulumi/pulumi's transpired-examples. If there are diffs, we open a PR against pulumi/pulumi and tag platform for review.
> 
> Assuming that the ☝️ is correct, this looks good.
> 
> P.S. We all got an email that pulumi/platform was deleted. Was that resolved? If not this will tag a non-existent team for review.

@iwahbe yup, that is the idea. It looks like pulumi/platform has been restored

.github/workflows/sync.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  convert-test-sync:
+    name: Sync PCL test files
+    runs-on: macos-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Run GitHub File Sync
+        uses: BetaHuhn/repo-file-sync-action@v1
```
AaronFriel, line 15: Supply chain security, pin this to ensure we don't run malicious code with our bot token:

```suggestion
        uses: BetaHuhn/repo-file-sync-action@40da4d517e0a744fce9e8aed03e63d6c8bafbdc0 # v1.17.21
```

.github/workflows/sync.yml
```diff
@@ -0,0 +1,18 @@
...
+        with:
+          GH_PAT: ${{ secrets.PULUMI_ACCESS_TOKEN }}
```
AaronFriel, line 17: Close! GH_PAT is asking for a GitHub personal access token, not a Pulumi access token.

This should work:

```suggestion
          GH_PAT: ${{ secrets.PULUMI_BOT_TOKEN }}
```

.github/workflows/sync.yml
```diff
@@ -0,0 +1,18 @@
+name: Sync Files
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  convert-test-sync:
+    name: Sync PCL test files
+    runs-on: macos-latest
```
AaronFriel, line 10: We have more Ubuntu runner capacity than macOS

```suggestion
    runs-on: ubuntu-latest
```

AaronFriel (changes requested):
Left some changes, this is an elegant approach. Happy to use a GH action here rather than maintain a shell script.

.github/workflows/sync.yml
aq17: 😬

AaronFriel (approved):
lgtm!

aq17:
bors merge

PR #366, opened 2022-10-12
Remove pu/pu codegen tests

aq17:


iwahbe (commented):
I'd prefer to hold off on this until https://github.com/pulumi/pulumi-yaml/pull/365 lands.

iwahbe (approved):
I'd prefer we not merge until the removed code is landed under test in pu/pu. That being said, it looks good here.

PR #370, opened 2022-10-17
Fix PCL sync workflow

aq17:
Fixes:
-directory path was incorrect

Maybe:
-update to GitHub PAT with full repo access


iwahbe (approved):
I see the fix to the source path. I don't see anything that touches which PAT we are using.

PR #379, opened 2022-10-22
Recognize the new core project-level config block

### LINKED ISSUE #369: Update pulumi-yaml to recognize the new core config block
https://github.com/pulumi/pulumi/pull/10832 adds a new core config block to Pulumi.yaml that the engine understands as defining project config. We should update pulumi-yaml to recognize and use that block rather than defining it's own "configuration" block.

---

As YAML is a language host using the Go SDK, it should rely on on the Go SDK config package (`GetConfig` and `IsConfigSecret`) to derive configuration keys, and use the structure of the values to infer their types, leaving validation up to the engine. **If that's not possible,** we should should use shared code from the engine to parse the `type` and `items` fields.

As a consequence of using `GetConfig`, it may not be possible to statically analyze a Pulumi program absent a stack file to determine if the a key is defined. We may want an enhancement to the Go SDK's config package to allow us to query for the _existence_ of a key or for static analysis to rely on the engine's parsing of the project file to be consistent.

As this would be a breaking change for `configuration` declarations; we should support the old key with a warning to avoid breaking users prior to GA.

aq17:
Fixes #369

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -129,6 +129,10 @@ func LoadYAMLBytes(filename string, source []byte) (*ast.TemplateDecl, syntax.Di
 	if tdiags.HasErrors() {
 		return nil, diags, nil
 	}
+	// TODO: warn if using old configuration block
+	// if t.Configuration.Entries != nil {
+	// 	diags = append(diags, syntax.Warning(nil, "Pulumi.yaml: root-level `configuration` field is deprecated; please use `config` instead.", ""))
+	// }
 
 	return t, diags, nil
 }
@@ -285,10 +289,12 @@ func (r *Runner) setDefaultProviders() error {
 }
 
 // PrepareTemplate prepares a template for converting or running
-func PrepareTemplate(t *ast.TemplateDecl, loader PackageLoader) (*Runner, syntax.Diagnostics, error) {
-	// set up context-free runner
-	r := newRunner(t, loader)
-
+func PrepareTemplate(t *ast.TemplateDecl, r *Runner, loader PackageLoader) (*Runner, syntax.Diagnostics, error) {
+	// If running a template also, we need to pass a runner through, since setting intermediates
+	// requires config via the pulumi Context
+	if r == nil {
+		r = newRunner(t, loader)
+	}
 	// runner hooks up default providers
 	err := r.setDefaultProviders()
 	if err != nil {
@@ -305,8 +311,14 @@ func PrepareTemplate(t *ast.TemplateDecl, loader PackageLoader) (*Runner, syntax
 }
 
 // RunTemplate runs the programEvaluator against a template using the given request/settings.
-func RunTemplate(ctx *pulumi.Context, t *ast.TemplateDecl, loader PackageLoader) error {
-	r, diags, err := PrepareTemplate(t, loader)
+func RunTemplate(ctx *pulumi.Context, t *ast.TemplateDecl, config map[string]string, loader PackageLoader) error {
+	r := newRunner(t, loader)
+	r.setIntermediates(config)
+	if r.sdiags.HasErrors() {
+		return &r.sdiags
+	}
+
+	r, diags, err := PrepareTemplate(t, r, loader)
 	if diags.HasErrors() {
 		return diags
 	}
@@ -616,18 +628,17 @@ func (e *programEvaluator) errorf(expr ast.Expr, format string, a ...interface{}
 }
 
 func (e programEvaluator) EvalConfig(r *Runner, node configNode) bool {
-
 	ctx := r.newContext(node)
 	c, ok := e.registerConfig(node)
 	if !ok {
-		e.config[node.Key.Value] = poisonMarker{}
-		msg := fmt.Sprintf("Error registering config [%v]: %v", node.Key.Value, ctx.sdiags.Error())
+		e.config[node.key().Value] = poisonMarker{}
+		msg := fmt.Sprintf("Error registering config [%v]: %v", node.key().Value, ctx.sdiags.Error())
 		err := e.pulumiCtx.Log.Error(msg, &pulumi.LogArgs{}) //nolint:errcheck
 		if err != nil {
 			return false
 		}
 	} else {
-		e.config[node.Key.Value] = c
+		e.config[node.key().Value] = c
 	}
 	return true
 }
@@ -685,7 +696,59 @@ func (r *Runner) Evaluate(ctx *pulumi.Context) syntax.Diagnostics {
 	return r.Run(programEvaluator{evalContext: eCtx, pulumiCtx: ctx})
 }
 
+func getPulumiConfNodes(config map[string]string) ([]configNode, error) {
+	nodes := make([]configNode, len(config))
+	var errors multierror.Error
+	idx := 0
+	for k, v := range config {
+		// We default types to strings to avoid error cascades on mis-typed values.
+		typ := ctypes.String
+		var value interface{} = v
+		if v, t, err := getConfigNode(v); err == nil {
+			typ = t
+			value = v
+		} else {
+			errors.Errors = append(errors.Errors, err)
+		}
+		n := configNodeEnv{
+			Key:   k,
+			Value: value,
+			Type:  typ,
+		}
+		nodes[idx] = n
+		idx++
+	}
+	return nodes, errors.ErrorOrNil()
+}
+
+// setIntermediates is called for convert and runtime evaluation
+func (r *Runner) setIntermediates(config map[string]string) {
+	if r.intermediates != nil {
+		return
+	}
+
+	r.intermediates = []graphNode{}
+	confNodes, err := getPulumiConfNodes(config)
+	if err != nil {
+		r.sdiags.Extend(syntax.Error(nil, err.Error(), ""))
+		return
+	}
+	// Topologically sort the intermediates based on implicit and explicit dependencies
+	intermediates, rdiags := topologicallySortedResources(r.t, confNodes)
+	r.sdiags.Extend(rdiags...)
+	if rdiags.HasErrors() {
+		return
+	}
+	if intermediates != nil {
+		r.intermediates = intermediates
+	}
+}
+
+// ensureSetup is called at runtime evaluation
 func (r *Runner) ensureSetup(ctx *pulumi.Context) {
+	// Our tests need to set intermediates, even though they don't have runtime config
+	r.setIntermediates(nil)
+
 	cwd, err := os.Getwd()
 	if err != nil {
 		r.sdiags.Extend(syntax.Error(nil, err.Error(), ""))
@@ -703,18 +766,29 @@ func (r *Runner) ensureSetup(ctx *pulumi.Context) {
 		"stack":   stack,
 	}
 	r.cwd = cwd
+}
 
-	r.intermediates = []graphNode{}
+// getConfigNode retrieves a runtime value and type from a config node.
+func getConfigNode(v string) (interface{}, ctypes.Type, error) {
+	// scalar config values are represented as their go literals, while arrays and objects
+	// are represented as JSON.
+	if i, err := strconv.ParseInt(v, 10, 64); err == nil {
+		return i, ctypes.Int, nil
+	}
+	if b, err := strconv.ParseBool(v); err == nil {
+		return b, ctypes.Boolean, nil
+	}
+	if f, err := strconv.ParseFloat(v, 64); err == nil {
+		return f, ctypes.Number, nil
+	}
 
-	// Topologically sort the intermediates based on implicit and explicit dependencies
-	intermediates, rdiags := topologicallySortedResources(r.t)
-	r.sdiags.Extend(rdiags...)
-	if rdiags.HasErrors() {
-		return
+	var value interface{}
+	if err := json.Unmarshal([]byte(v), &value); err == nil {
+		typ, err := ctypes.TypeValue(value)
+		return value, typ, err
 	}
-	if intermediates != nil {
-		r.intermediates = intermediates
-	}
+
+	return v, ctypes.String, nil
 }
 
 func (r *Runner) Run(e Evaluator) syntax.Diagnostics {
@@ -739,7 +813,7 @@ func (r *Runner) Run(e Evaluator) syntax.Diagnostics {
 		switch kvp := kvp.(type) {
 		case configNode:
 			if ctx != nil {
-				err := ctx.Log.Debug(fmt.Sprintf("Registering config [%v]", kvp.Key.Value), &pulumi.LogArgs{})
+				err := ctx.Log.Debug(fmt.Sprintf("Registering config [%v]", kvp.key().Value), &pulumi.LogArgs{})
 				if err != nil {
 					return returnDiags()
 				}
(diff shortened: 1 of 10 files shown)


Frassle:
My understanding of @AaronFriel's explanation around this was that YAML was going to stop looking at the config block directly and look at config and configSecretKeys in the run request from the engine instead.

aq17:
So YAML would use the Go SDK to do things like `conf.TryInt("instanceSize")` instead of the project file config block? If so, would it make sense to have a separate syntax instead of interpolated expressions to get a config value, i.e. `conf.instanceSize`

Would we still need to read in the new config block while loading the template to get the default value of config values if not set?

cc @iwahbe

aq17:
> So YAML would use the Go SDK to do things like `conf.TryInt("instanceSize")` instead of the project file config block? If so, would it make sense to have a separate syntax instead of interpolated expressions to get a config value, i.e. `conf.instanceSize`

> cc @iwahbe

iwahbe:
We already use the go SDK to pull config values, see https://github.com/pulumi/pulumi-yaml/blob/0197f6544aee26c557ba84573cd0a1506eed3e3c/pkg/pulumiyaml/run.go#L845-L850

In the past, we needed to read the config to known what values to pull into scope, and what defaults to provide. If defaults will be handled for us, we can just inject config values into our scope and error (during [topologicallySortedResources](https://github.com/pulumi/pulumi-yaml/blob/0197f6544aee26c557ba84573cd0a1506eed3e3c/nil#L50)) when attempting to use missing values.

I don't see why changing the implementation should change the user syntax.

pkg/pulumiyaml/ast/template.go
```diff
@@ -96,6 +96,7 @@ func (d *ConfigMapDecl) parse(name string, node syntax.Node) syntax.Diagnostics
 	if !ok {
 		return syntax.Diagnostics{syntax.NodeError(node, fmt.Sprintf("%v must be an object", name), "")}
 	}
+	// need to support project config shorthand, i.e. "instanceCount: 5"
```
iwahbe, line 99: This can probably go now.

pkg/pulumiyaml/run.go
```diff
@@ -781,62 +781,69 @@ func (r *Runner) Run(e Evaluator) syntax.Diagnostics {
...
-			"Cannot mark a configuration value as not secret"+
-				" if the associated config value is secret")
+		if isSecretInConfig && c.Secret != nil && !c.Secret.Value {
+			return e.error(c.Secret,
+				"Cannot mark a configuration value as not secret"+
+					" if the associated config value is secret")
+		}
+	case configNodeEnv:
+		k = intm.Key
+		expectedType = intm.Type
```
iwahbe, line 846: If we never expect to get other node types, we should make sure they don't slip in. If we do, a comment explaining that would be awesome.
```suggestion
		expectedType = intm.Type
	default:
		return e.errorf(nil, "internal error: unexpected config type %T", intm)
```

pkg/pulumiyaml/run.go
```diff
@@ -911,16 +918,20 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
...
 	// config section. We need to wrap it in `pulumi.ToSecret`.
-	if (c.Secret != nil && c.Secret.Value) && !isSecretInConfig {
-		v = pulumi.ToSecret(v)
+	if intm, ok := intm.(configNodeYaml); ok {
+		c := intm.Value
+		if (c.Secret != nil && c.Secret.Value) && !isSecretInConfig {
+			v = pulumi.ToSecret(v)
+		}
 	}
+
```
iwahbe, line 934: I think there is an env var that is used to mark `PULUMI_CONFIG` values as secret. We should address that here.

pkg/pulumiyaml/sort.go
```diff
@@ -37,16 +40,44 @@ func (e variableNode) key() *ast.StringExpr {
 	return e.Key
 }
 
-type configNode ast.ConfigMapEntry
+type configNode interface {
+	valueKind() string
+	key() *ast.StringExpr
```
iwahbe, line 45: All this really means is we want `configNode` to implement `graphNode`. We can say that more directly:
```suggestion
	graphNode
```

pkg/pulumiyaml/sort.go
```diff
@@ -73,6 +104,58 @@ func topologicallySortedResources(t *ast.TemplateDecl) ([]graphNode, syntax.Diag
 			sorted = append(sorted, node)
 		}
 	}
+	var projConf map[string]interface{}
+	var projConfStr = os.Getenv("PULUMI_CONFIG")
+	if err := json.Unmarshal([]byte(projConfStr), &projConf); err == nil {
```
iwahbe, line 109: If there is an error unmarshalling, we need to report it. We should almost never swallow an error.

pkg/pulumiyaml/sort.go
```diff
@@ -73,6 +104,58 @@ func topologicallySortedResources(t *ast.TemplateDecl) ([]graphNode, syntax.Diag
 			sorted = append(sorted, node)
 		}
 	}
+	var projConf map[string]interface{}
```
iwahbe, line 107: Instead of grabbing the config in `topologicallySortedResources`, could we adjust the signature to accept some additional config nodes.
```go
func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNode) ([]graphNode, syntax.Diagnostics)
```

We should be able to handle all config nodes the same way, replacing `t.Config.Entries` with `append(extraConfig, templateConfig)` in the above `for` loop.

Here `templateConfig` is simply:
```go
templateConfig := make([]configNode, len(t.Configuration.Entries)
for i, kip := range t.Configuration.Entries {
    templateConfig[i] = configNode(kvp)
}
```

This gives us two benefits:
1. DRY: We always handle adding `configNode`s the same way.
2. `topologicallySortedResources` remains a pure function, making it much easier to test and maintain. We can put the `PULUMI_CONFIG -> []configNode` logic in a separate function.

iwahbe (changes requested):

pkg/pulumiyaml/run.go
aq17: I'm not quite sure if Friel wanted to add `PULUMI_CONFIG_SECRET_KEYS` (https://github.com/pulumi/pulumi-yaml/issues/369) or if it already exists because I can't find any documentation on it..

pkg/pulumiyaml/run.go
```diff
@@ -685,6 +685,37 @@ func (r *Runner) Evaluate(ctx *pulumi.Context) syntax.Diagnostics {
 	return r.Run(programEvaluator{evalContext: eCtx, pulumiCtx: ctx})
 }
 
+func (r *Runner) getPulumiConfNodes() []configNode {
+	var projConf map[string]interface{}
+	var projConfStr = os.Getenv("PULUMI_CONFIG")
+	nodes := make([]configNode, 0)
+
+	if err := json.Unmarshal([]byte(projConfStr), &projConf); err != nil {
```
iwahbe, line 693: We shouldn't swallow errors here. There are many reasons that `err != nil` beyond `projConfStr` being empty. We can actually check if it is empty; see above.

pkg/pulumiyaml/run.go
```diff
@@ -685,6 +685,37 @@ func (r *Runner) Evaluate(ctx *pulumi.Context) syntax.Diagnostics {
 	return r.Run(programEvaluator{evalContext: eCtx, pulumiCtx: ctx})
 }
 
+func (r *Runner) getPulumiConfNodes() []configNode {
+	var projConf map[string]interface{}
+	var projConfStr = os.Getenv("PULUMI_CONFIG")
```
iwahbe, line 690:
```suggestion
	projConfStr, ok := os.LookupEnv("PULUMI_CONFIG")
	if !ok {
		return nil
	}
```

pkg/pulumiyaml/run.go
```diff
@@ -685,6 +685,37 @@ func (r *Runner) Evaluate(ctx *pulumi.Context) syntax.Diagnostics {
...
+	nodes := make([]configNode, 0)
```
iwahbe, line 691: If were going to use the `make([]type, length)` syntax here, we can use it to ensure a single allocation by moving it after we deserialize `projConfStr`:
```suggestion
	nodes := make([]configNode, len(projConf))
```

pkg/pulumiyaml/run.go
```diff
@@ -911,16 +951,23 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
...
-	if (c.Secret != nil && c.Secret.Value) && !isSecretInConfig {
-		v = pulumi.ToSecret(v)
+	switch intm := intm.(type) {
+	case configNodeYaml:
+		c := intm.Value
+		if (c.Secret != nil && c.Secret.Value) && !isSecretInConfig {
+			v = pulumi.ToSecret(v)
+		}
+	case configNodeEnv:
+
```
iwahbe, line 968: What goes here?

iwahbe (approved):
It would be great to start seeing tests for `getPulumiConfNodes`

iwahbe (changes requested):
Misclicked on approve

aq17:
For updating our [examples](https://github.com/pulumi/pulumi-yaml/tree/main/examples) and being able to convert YAML->PCL, we may need to parse the `config` block? @iwahbe

iwahbe:
> For updating our [examples](https://github.com/pulumi/pulumi-yaml/tree/main/examples) and being able to convert YAML->PCL, we may need to parse the `config` block? @iwahbe

Parsing the config block is complicated already, and will get more so when we introduce new sources of config. Even now, consider the following 3 files:

Pulumi.yaml:
```
name: proj
runtime: yaml
main: src/Main.yaml
config:
  prefix: 
    default: abc
```

Pulumi.dev.yaml:
```
config:
  pros:tag: tag2
```

src/Main.yaml:
```
resources:
  bucket:
    type: aws:s3:Bucket
    properties:
      bucket: ${prefix}-bucket
      tags:
        stackTag: ${tag}
```

I can see how we will need the information though. Right now, I think we can continue going through `configuration` for this. In the future we can add the ability to describe the config when converting YAML->PCL.

Frassle:
> In the future we can add the ability to describe the config when converting YAML->PCL.

I think that might sync well with the ideas of code generating config types for the other languages. We should try to make sure those two things use the same language.

pkg/pulumiyaml/run.go
```diff
@@ -285,10 +289,11 @@ func (r *Runner) setDefaultProviders() error {
 }
 
 // PrepareTemplate prepares a template for converting or running
-func PrepareTemplate(t *ast.TemplateDecl, loader PackageLoader) (*Runner, syntax.Diagnostics, error) {
-	// set up context-free runner
-	r := newRunner(t, loader)
-
+func PrepareTemplate(t *ast.TemplateDecl, r *Runner, loader PackageLoader) (*Runner, syntax.Diagnostics, error) {
```
iwahbe, line 292: I think there needs to be a comment explaining why you would/would not pass a runner through.

pkg/pulumiyaml/run.go
```diff
@@ -318,7 +329,7 @@ func RunTemplate(ctx *pulumi.Context, t *ast.TemplateDecl, loader PackageLoader)
 	}
 
 	// runtime evaluation here
-	diags.Extend(r.Evaluate(ctx)...)
+	diags.Extend(r.Evaluate(ctx, config)...)
```
iwahbe, line 332: Why does `Evaluate` need a copy of `config` when the config has already been added to the runner with `r.setIntermediates` (L315)?

pkg/pulumiyaml/run.go
```diff
@@ -571,7 +582,8 @@ type Evaluator interface {
 
 type programEvaluator struct {
 	*evalContext
-	pulumiCtx *pulumi.Context
+	pulumiCtx     *pulumi.Context
+	runtimeConfig map[string]string
```
iwahbe, line 588: Both `Runner` and `programEvaluator` have a config option? I think it would be preferred if we could stick with just `Runner` unless they do different things, if so how are they different?

pkg/pulumiyaml/run.go
```diff
@@ -680,12 +691,48 @@ func (e programEvaluator) EvalOutput(r *Runner, node ast.PropertyMapEntry) bool
 	return true
 }
 
-func (r *Runner) Evaluate(ctx *pulumi.Context) syntax.Diagnostics {
+func (r *Runner) Evaluate(ctx *pulumi.Context, config map[string]string) syntax.Diagnostics {
```
iwahbe, line 696: See above comment:
> Why does Evaluate need a copy of config when the config has already been added to the runner with r.setIntermediates (L315)?

pkg/pulumiyaml/run.go
```diff
@@ -680,12 +691,48 @@ func (e programEvaluator) EvalOutput(r *Runner, node ast.PropertyMapEntry) bool
...
+			Value: v,
+			Type:  getConfigType(v),
+		}
+		nodes[idx] = n
+		idx++
+	}
+	return nodes, nil
+}
+
+// ensureSetupCtxless is called for convert and runtime evaluation
```
iwahbe, line 716: This comment refers to an old name.

pkg/pulumiyaml/run.go
```diff
@@ -680,12 +691,48 @@ func (e programEvaluator) EvalOutput(r *Runner, node ast.PropertyMapEntry) bool
...
+	}
+	if intermediates != nil {
+		r.intermediates = intermediates
+	}
+	return nil
 }
 
-func (r *Runner) ensureSetup(ctx *pulumi.Context) {
+// ensureSetup is called at runtime evaluation
+func (r *Runner) ensureSetup(ctx *pulumi.Context, config map[string]string) {
```
iwahbe, line 743: `config` is never used.

pkg/pulumiyaml/run.go
```diff
@@ -703,28 +750,31 @@ func (r *Runner) ensureSetup(ctx *pulumi.Context) {
...
+}
 
-	r.intermediates = []graphNode{}
-
-	// Topologically sort the intermediates based on implicit and explicit dependencies
-	intermediates, rdiags := topologicallySortedResources(r.t)
-	r.sdiags.Extend(rdiags...)
-	if rdiags.HasErrors() {
-		return
+func getConfigType(v string) ctypes.Type {
```
iwahbe, line 765: This doesn't work for array types. I'll open a PR into your branch that will fix this.

iwahbe (commented):
I left some comments. There are a couple places to clean up, but I think its getting close.

pkg/pulumiyaml/run.go
aq17: Leftover code, thanks for catching

pkg/pulumiyaml/run.go
```diff
@@ -366,6 +378,8 @@ type Runner struct {
 
 	sdiags syncDiags
 
+	intermediatesSet bool
```
iwahbe, line 381: This field is not strictly necessary, since we can test `intermediates == nil`. I believe thats what we did in the past.

iwahbe (approved):
I left 1 very minor nit, but LGTM!

PR #382, opened 2022-10-27
Escaped interpolated strings remove one more $ sign from evaluation

Zaid-Ajaj:
The following YAML program:
```yaml
name: yamlescape
runtime: yaml
outputs:
  message: hello $${world}
```
evaluated `message` to just `hello $${world}` where it should have been `hello ${world}`

It turns out, the parsed Value node of `hello $${world}` is just `StringExpr` where it should have been an `InterpolatedExpr` so I've added interpolation logic where we construct `StringExpr` (I have a feeling it should be somewhere else?) 

I've also changed the escaping logic from `strings.ReplaceAll(p.Text, "$", "$$")` to `strings.ReplaceAll(p.Text, "$$", "$")` such that `$$` now is replaced with `$`

### DIFF
--- a/pkg/pulumiyaml/ast/expr.go
+++ b/pkg/pulumiyaml/ast/expr.go
@@ -117,17 +117,12 @@ func (x *StringExpr) GetValue() string {
 
 // StringSyntax creates a new string literal expression with the given value and associated syntax.
 func StringSyntax(node *syntax.StringNode) *StringExpr {
-	parts, diags := parseInterpolate(node, node.Value())
-	if diags.HasErrors() {
-		return &StringExpr{exprNode: expr(node), Value: node.Value()}
-	}
-
-	if len(parts) == 1 && parts[0].Value == nil {
-		interpolated := &InterpolateExpr{Parts: parts}
-		return &StringExpr{exprNode: expr(node), Value: interpolated.String()}
-	}
-
 	return &StringExpr{exprNode: expr(node), Value: node.Value()}
+}
+
+// StringSyntaxValue creates a new string literal expression with the given syntax and value.
+func StringSyntaxValue(node *syntax.StringNode, value string) *StringExpr {
+	return &StringExpr{exprNode: expr(node), Value: value}
 }
 
 // String creates a new string literal expression with the given value.
@@ -303,7 +298,7 @@ func ParseExpr(node syntax.Node) (Expr, syntax.Diagnostics) {
 			case 1:
 				switch {
 				case interpolate.Parts[0].Value == nil:
-					return StringSyntax(node), diags
+					return StringSyntaxValue(node, interpolate.Parts[0].Text), diags
 				case interpolate.Parts[0].Text == "":
 					return &SymbolExpr{
 						exprNode: expr(node),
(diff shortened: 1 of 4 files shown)


iwahbe:
> It turns out, the parsed Value node of `hello $${world}` is just `StringExpr` where it should have been an `InterpolatedExpr` so I've added interpolation logic where we construct `StringExpr` (I have a feeling it should be somewhere else?)

I don't agree that it should be an `InterpolateExpr`, since we don't want to interpolate anything. Its still a string literal that happens to have an escaped `$` in it.

> I've also changed the escaping logic from `strings.ReplaceAll(p.Text, "$", "$$")` to `strings.ReplaceAll(p.Text, "$$", "$")` such that `$$` now is replaced with `$`

I don't understand why this needs to change in light of my above comment.

iwahbe (changes requested):
The tests are great. I think the fix can be much simpler though. 

```patch
 pkg/pulumiyaml/ast/expr.go | 9 +++++++--
 1 file changed, 7 insertions(+), 2 deletions(-)

diff --git a/pkg/pulumiyaml/ast/expr.go b/pkg/pulumiyaml/ast/expr.go
index 066f443..113a128 100644
--- a/pkg/pulumiyaml/ast/expr.go
+++ b/pkg/pulumiyaml/ast/expr.go
@@ -117,7 +117,12 @@ func (x *StringExpr) GetValue() string {
 
 // StringSyntax creates a new string literal expression with the given value and associated syntax.
 func StringSyntax(node *syntax.StringNode) *StringExpr {
-	return &StringExpr{exprNode: expr(node), Value: node.Value()}
+	return StringSyntaxValue(node, node.Value())
+}
+
+// StringSyntaxValue creates a new string literal expression with the given syntax and value.
+func StringSyntaxValue(node *syntax.StringNode, value string) *StringExpr {
+	return &StringExpr{exprNode: expr(node), Value: value}
 }
 
 // String creates a new string literal expression with the given value.
@@ -293,7 +298,7 @@ func ParseExpr(node syntax.Node) (Expr, syntax.Diagnostics) {
 			case 1:
 				switch {
 				case interpolate.Parts[0].Value == nil:
-					return StringSyntax(node), diags
+					return StringSyntaxValue(node, interpolate.Parts[0].Text), diags
 				case interpolate.Parts[0].Text == "":
 					return &SymbolExpr{
 						exprNode: expr(node),
```

Zaid-Ajaj:
Ah yes, it was in `ParseExpr` 🙌 that's what I was looking for, escaping is done once in `parseInterpolate` so need to apply the escaping in multiple places.

pkg/pulumiyaml/ast/expr.go
```diff
@@ -141,7 +146,7 @@ type InterpolateExpr struct {
 func (n *InterpolateExpr) String() string {
 	var str strings.Builder
 	for _, p := range n.Parts {
-		str.WriteString(strings.ReplaceAll(p.Text, "$", "$$"))
+		str.WriteString(strings.ReplaceAll(p.Text, "$$", "$"))
```
iwahbe, line 149: InterpolateExpr.String is going from the runtime representation back to the textual representation. We should be re-escaping here, not double unescaping.

Zaid-Ajaj: I reverted the change in this function to the original implementation and added a comment explaining what it was doing ✅

iwahbe (approved):
LGTM

PR #383, opened 2022-10-27
Ensure that node order is deterministic

iwahbe:
We were seeing non-deterministic test results such as https://github.com/pulumi/pulumi-yaml/actions/runs/3340731857/attempts/1. The problem was inconsistent node order when the walker was running. 

Since the top-sort algorithm does not rely on iteration order, we are free to choose an arbitrary order. I have chosen insertion order following the principle of least surprise.

The problem and the solution can be verified by running the `pulumiyaml/TestVersionConflicts` tests with a high count number. You will see intermittent failures before the change, but not after.

### DIFF
--- a/pkg/pulumiyaml/sort.go
+++ b/pkg/pulumiyaml/sort.go
@@ -56,6 +56,28 @@ func topologicallySortedResources(t *ast.TemplateDecl) ([]graphNode, syntax.Diag
 
 	// Precompute dependencies for each resource
 	intermediates := map[string]graphNode{}
+	// The list of keys to intermediates in the order they were first added.
+	sortedIntermediatesKeys := []string{}
+
+	// Add a new node to intermediates.
+	addItermediate := func(key string, node graphNode) {
+		_, duplicate := intermediates[key]
+		intermediates[key] = node
+		if !duplicate {
+			sortedIntermediatesKeys = append(sortedIntermediatesKeys, key)
+		}
+	}
+
+	// A list of graphNodes from intermediates in the order they were inserted in.
+	// This ensures iteration order is deterministic.
+	intermediateValues := func() []graphNode {
+		sorted := make([]graphNode, len(sortedIntermediatesKeys))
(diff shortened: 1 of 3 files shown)


aq17 (approved):

pkg/pulumiyaml/sort.go
```diff
@@ -56,6 +56,28 @@ func topologicallySortedResources(t *ast.TemplateDecl) ([]graphNode, syntax.Diag
 
 	// Precompute dependencies for each resource
 	intermediates := map[string]graphNode{}
+	// The list of keys to intermediates in the order they were first added.
+	sortedIntermediatesKeys := []string{}
+
+	// Add a new node to intermediates.
+	addItermediate := func(key string, node graphNode) {
```
kpitzen, line 63:
```suggestion
	addIntermediate := func(key string, node graphNode) {
```

pkg/pulumiyaml/sort.go
```diff
@@ -81,7 +103,7 @@ func topologicallySortedResources(t *ast.TemplateDecl) ([]graphNode, syntax.Diag
 		diags = append(diags, cdiags...)
 
 		if !cdiags.HasErrors() {
-			intermediates[rname] = node
+			addItermediate(rname, node)
```
kpitzen, line 106:
```suggestion
			addIntermediate(rname, node)
```

pkg/pulumiyaml/sort.go
```diff
@@ -93,7 +115,7 @@ func topologicallySortedResources(t *ast.TemplateDecl) ([]graphNode, syntax.Diag
 		diags = append(diags, cdiags...)
 
 		if !cdiags.HasErrors() {
-			intermediates[vname] = node
+			addItermediate(vname, node)
```
kpitzen, line 118:
```suggestion
			addIntermediate(vname, node)
```

kpitzen (approved):
Other than the small typo, LGTM!

PR #387, opened 2022-10-31
Parse complex config types

iwahbe:
This PR adds error handling for array based types. 

It also makes `configNode.value()` return the actual value instead of its string representation, but that isn't currently used.


aq17 (approved):
Looks great 🎉

PR #389, opened 2022-11-01
Make resource ref "id" insertion type aware

### LINKED ISSUE #386: `pulumi convert` turns references to resources into `res.id` which is not valid in other languages
Creating the default `pulumi new serverless-aws-yaml` and then `pulumi convert --language typescript --out ./ts` succeeds - but then fails to typescheck in TS with:

```console
reviewing update (dev)

View Live: https://app.pulumi.com/lukehoban/demo2/dev/previews/158d94d9-3fcb-4d19-871c-0bf40984b85b

     Type                 Name       Plan     Info
     pulumi:pulumi:Stack  demo2-dev           1 error
 
Diagnostics:
  pulumi:pulumi:Stack (demo2-dev):
    error: Running program '/Users/lukehoban/dd/ced2022demos/demo2/demo2ts' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    index.ts(32,9): error TS2322: Type 'Output<string>' is not assignable to type 'Input<Function> | undefined'.
      Type 'Output<string>' is not assignable to type 'OutputInstance<Function>'.
        Types of property 'apply' are incompatible.
          Type '{ <U>(func: (t: string) => Promise<U>): Output<U>; <U>(func: (t: string) => OutputInstance<U>): Output<U>; <U>(func: (t: string) => U): Output<...>; }' is not assignable to type '{ <U>(func: (t: Function) => Promise<U>): Output<U>; <U>(func: (t: Function) => OutputInstance<U>): Output<U>; <U>(func: (t: Function) => U): Output<...>; }'.
            Types of parameters 'func' and 'func' are incompatible.
              Types of parameters 't' and 't' are incompatible.
                Type 'string' is not assignable to type 'Function'. 
```

Similar code is generated in other languages - which will similarly not typecheck correctly for them.

The issue is that this:

```yaml
eventHandler: ${fn}
```

Becomes this:

```ts
eventHandler: fn.id,
```

But should be this:

```ts
eventHandler: fn,
```

iwahbe:
Fixes #386

Pulumi YAML used to turn resource refs into strings by adding the "id" property to the end of resources. Sometimes this is the correct behavior, reflecting Pulumi YAMLs ability to pass resources where strings go by converting them to strings vie their ID.

In other places, the resource itself needs to be passed directly, and the implicit string conversion is not OK.

This PR makes the ID insertion conditional on the receiver being a string type. It is tested by our own examples, but I have confirmed that it works on the repro introduced in the PR.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -29,3 +29,6 @@
 
 - Escaped interpolated strings now remove one extra dollar sign.
   [#382](https://github.com/pulumi/pulumi-yaml/pull/382)
+
+- Only insert "id" in ejected resource refs when the receiver type is a string.
+  [#389](https://github.com/pulumi/pulumi-yaml/pull/389)
--- a/pkg/pulumiyaml/codegen/load.go
+++ b/pkg/pulumiyaml/codegen/load.go
@@ -27,7 +27,6 @@ type importer struct {
 	referencedStacks []string
 
 	loader          pulumiyaml.PackageLoader
-	typing          pulumiyaml.Typing
 	configuration   map[string]*model.Variable
 	variables       map[string]*model.Variable
 	stackReferences map[string]*model.Variable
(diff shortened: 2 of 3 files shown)


aq17 (approved):
LGTM!

PR #391, opened 2022-11-01
Swallow required plugins error

aq17:
Related to #369

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -89,7 +89,10 @@ func (host *yamlLanguageHost) GetRequiredPlugins(ctx context.Context,
 	pkgs, pluginDiags := pulumiyaml.GetReferencedPlugins(template)
 	diags.Extend(pluginDiags...)
 	if diags.HasErrors() {
-		return nil, diags
+		// We currently swallow the error to allow runtime config to evaluate
+		// Specifically, if one sets a config key via the CLI but not within the `config` block
+		// of their YAML program, it would error.
+		return &pulumirpc.GetRequiredPluginsResponse{}, nil
 	}
 	var plugins []*pulumirpc.PluginDependency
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
Thanks for leaving a comment explaining what's going on. LGTM!

pkg/server/server.go
```diff
@@ -89,7 +89,10 @@ func (host *yamlLanguageHost) GetRequiredPlugins(ctx context.Context,
 	pkgs, pluginDiags := pulumiyaml.GetReferencedPlugins(template)
 	diags.Extend(pluginDiags...)
 	if diags.HasErrors() {
-		return nil, diags
+		// We currently swallow the error to allow project config to evaluate
+		// Specifically, if one sets a config key via the CLI but not within the `config` block
```
Frassle, line 93: This seems odd? Why does yaml program parsing require stack config at all?

aq17: It doesn't require stack config specifically, the issue is we don't have a config node for project config until runtime. So when it's parsing and sees something like `${myConfigKey}`, where we don't have `myConfigKey` under the `config` block, it would treat it as an unknown resource and error

iwahbe: Determining referenced plugins requires analyzing what providers are created, if they are set to default, if resources specify explicit providers, ect. This is all done after we perform the top-sort on nodes. Missing config nodes are caught by top-sort and we error out. This error would prevent the rest of the `pulumi up`, even when the update itself is valid.

Frassle: This sounds like a TODO to fix, is there an issue for it?

Frassle: Oh I see theres already https://github.com/pulumi/pulumi-yaml/pull/392 on it

PR #392, opened 2022-11-01
Assume unknown key is project config key on convert

aq17:


aq17:
@iwahbe  still looking into this right now, this gets past `PrepareTemplate` but not `ImportTemplate` -> `importExpr` yet

pkg/pulumiyaml/sort.go
```diff
@@ -234,3 +242,7 @@ func checkUniqueNode(intermediates map[string]graphNode, node graphNode) syntax.
 	}
 	return diags
 }
+
+func isPotentialConfig(name string) bool {
+	return len(strings.Split(name, ":")) == 2
```
iwahbe, line 247: This is simpler and faster.
```suggestion
	return strings.Count(name, ":") == 2
```

pkg/pulumiyaml/sort.go
```diff
@@ -166,11 +167,18 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 
 		e, ok := intermediates[name.Value]
 		if !ok {
-			diags.Extend(ast.ExprError(name, fmt.Sprintf("resource %q not found", name.Value), ""))
-			return false
+			if isPotentialConfig(name.Value) {
```
iwahbe, line 170: We will always need to generate the diagnostic. The difference is we will inject a dummy node instead of stopping the algorithm. The caller of `topologicallySortedResources` can then choose to ignore diags (which we will do for eject and get plugins, but nothing else).

aq17:
Closed in favor of https://github.com/pulumi/pulumi-yaml/pull/393

PR #395, opened 2022-11-02
Set version as not current on sync

aq17:
We want to set `PULUMI_ACCEPT=true` to rerun program generation on synced examples in `pulumi/pulumi` https://github.com/pulumi/pulumi/blob/master/.github/workflows/ci-run-test.yml#L300


iwahbe (approved):
🚢 

It would be great to add a note in the PR body saying "This PR was run with `PULUMI_ACCEPT=true`. You need to manually confirm that changes are valid before accepting.

PR #407, opened 2022-11-11
Parse config block for pulumi convert

### LINKED ISSUE #406: Parse `config` block for `pulumi convert`
To properly support `pulumi convert`, the `config` block must be parsed to collect any type info for each config key

aq17:
In order to properly handle `pulumi convert`, we need to parse the project-level `config` block and fetch and type info for keys
Fixes #406


pkg/pulumiyaml/ast/template.go
```diff
@@ -102,16 +102,27 @@ func (d *ConfigMapDecl) parse(name string, node syntax.Node) syntax.Diagnostics
...
-		var v *ConfigParamDecl
-		vname := fmt.Sprintf("%s.%s", name, kvp.Key.Value())
-		vdiags := parseField(vname, reflect.ValueOf(&v).Elem(), kvp.Value)
-		diags.Extend(vdiags...)
-
-		entries[i] = ConfigMapEntry{
-			syntax: kvp,
-			Key:    StringSyntax(kvp.Key),
-			Value:  v,
+		if _, ok := kvp.Value.(*syntax.ObjectNode); !ok {
```
aq17, line 105: @AaronFriel `pulumi convert` was already calling `RewriteShorthandConfigValues()`, turning something like `config: aws:region: us-west-1` into `aws:region: value: us-west-1`; however, the node was still interpreted as its direct value/ string node here, not an object node. When we deprecate `configuration` I think we'd loosen the constraint here anyways

pkg/server/server.go
```diff
@@ -157,9 +158,17 @@ func (host *yamlLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest
...
 		}
 		defer loader.Close()
-
+		conf := make(map[string]string)
+		// Strip the project prefix
+		projPrefix := pctx.Project() + ":"
+		for k, v := range req.GetConfig() {
+			if strings.HasPrefix(k, projPrefix) {
+				k = strings.Split(k, projPrefix)[1]
+			}
```
iwahbe, line 167: There is a built in function that does this:
```suggestion
			k = strings.TrimPrefix(k, projPrefix)
```

iwahbe (commented):
This works when types line up. I think we need to add a check for incompatible types between configs though:

Consider the following example:
```yaml
name: test
runtime: yaml
config:
  foo:
    type: string
    default: "true"
configuration:
  foo:
    type: boolean
outputs:
  foo: ${foo}
```

Without any stack config set, running `pulumi up` here works. If you run `pulumi config set foo bar` and run `pulumi up` again, you get a runtime conversion error exposing go internals:
```
    error: strconv.ParseBool: parsing "bar": invalid syntax
    
      on Pulumi.yaml line 9:
       9:   foo:
    error: Error registering config [foo]: no diagnostics
```

Unfortunately, thats the best error you'll get.
If we change `configuration.foo.type` to `list<boolean>`, the error becomes truly terrible:

```
   error: invalid character 'b' looking for beginning of value
    
      on Pulumi.yaml line 8:
       8:   foo:
    error: Error registering config [foo]: no diagnostics
```

I think we should take the opportunity to assert that `config` types agree with `configuration` types.

pkg/pulumiyaml/run.go
```diff
@@ -1507,6 +1503,15 @@ func (e *programEvaluator) evaluatePropertyAccess(expr ast.Expr, access *ast.Pro
...
 		receiver = v
+	} else if v, ok := e.config[stripConfigNamespace(resourceName)]; ok {
+		receiver = v
+		e.addDiag(&syntax.Diagnostic{
+			Diagnostic: hcl.Diagnostic{
+				Severity: hcl.DiagWarning,
+				Summary:  "Project config keys should not be namespaced;",
+				Detail:   fmt.Sprintf("Use %s instead of %s", stripConfigNamespace(resourceName), resourceName),
+			},
+		})
```
AaronFriel, line 1514: I don't think this is enforced by Pulumi for other languages - what's the motivation for us doing that here?

aq17: No firm reason – we can remove the warn to be consistent then

pkg/pulumiyaml/analyser.go
```diff
@@ -1083,6 +1083,17 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	if optional {
 		typ = &schema.OptionalType{ElementType: typ}
 	}
+	// check for incompatible types between config/ configuration
+	if typExisting, ok := tc.configuration[k]; ok {
+		if typExisting.String() != typ.String() {
+			ctx := r.newContext(node)
+			ctx.sdiags.diags.Extend(syntax.NodeError(node.key().Syntax(), fmt.Sprintf("config key cannot have conflicting types %v, %v",
+				codegen.UnwrapType(typExisting), codegen.UnwrapType(typ)), ""))
```
iwahbe, line 1091: I think this can be deleted. It's a duplicate of the next `ctx.error(...)` call on the next line.

pkg/pulumiyaml/analyser.go
```diff
@@ -1083,6 +1083,17 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
...
+			ctx.error(nil, fmt.Sprintf("config key %s cannot have conflicting types %v, %v",
```
iwahbe, line 1092: We have ` ctx.errorf` method to abbreviate this:
```suggestion
			ctx.errorf(nil, "config key %s cannot have conflicting types %v, %v",
```

pkg/pulumiyaml/sort.go
```diff
@@ -178,9 +167,12 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 
 		e, ok := intermediates[name.Value]
 		if !ok {
-			diags.Extend(ast.ExprError(name, fmt.Sprintf("resource %q not found", name.Value), ""))
-			e = missingNode{name}
-			addIntermediate(name.Value, e)
+			if e2, ok := intermediates[stripConfigNamespace(name.Value)]; ok {
+				e = e2
+			} else {
+				diags.Extend(ast.ExprError(name, fmt.Sprintf("resource %q not found", name.Value), ""))
```
iwahbe, line 173: Can we change `resource` to `"resource, variable or config value"`?

pkg/pulumiyaml/sort.go
```diff
@@ -256,3 +248,11 @@ func isConfigNodeEnv(n graphNode) bool {
 	_, ok := n.(configNodeEnv)
 	return ok
 }
+
+func stripConfigNamespace(s string) string {
```
iwahbe, line 252: Does this mean that `aws:region` and `my-proj:region` are actually the same node? This seems scary to me.

pkg/server/server.go
```diff
@@ -157,9 +158,15 @@ func (host *yamlLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest
 			return err
 		}
 		defer loader.Close()
-
+		conf := make(map[string]string)
```
iwahbe, line 161: We may as well save the allocation.
```suggestion
		conf := make(map[string]string, len(req.GetConfig())
```

iwahbe (changes requested):
The only part that seriously concerns me is stripping config prefixs. The rest looks good with some nits.

pkg/pulumiyaml/sort.go
aq17: True, though we can't get the project name due to the lack of context. We could store the first prefix to hit this as in a var, and then if the same stripped key, check its prefix against the var and error if different?

iwahbe: I think we should have a way to have both `aws:region` and `my-proj:region`. We **need** to be able to express `aws:region` and `aws-native:region` at the same time.

aq17: Actually this path is specifically for `pulumi up` I believe, so we should have access to the context

pkg/pulumiyaml/sort.go
```diff
@@ -256,3 +248,11 @@ func isConfigNodeEnv(n graphNode) bool {
 	_, ok := n.(configNodeEnv)
 	return ok
 }
+
+func stripConfigNamespace(n, s string) string {
+	tok := strings.Split(s, ":")
+	if len(tok) == 2 && tok[0] == n {
+		return tok[1]
+	}
+	return s
```
iwahbe, line 257: AFAK config keys must be `SEG1:SEG2`. We will never have a scenario where `len(tok) > 2`. This lets us simplify this block to `return strings.TrimPrefix(n+":", s)`.

iwahbe (approved):
I think this feature should be tested. The implementation looks good though. After tests are added, LGTM!

pkg/pulumiyaml/codegen/prepare_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	host, err := newPluginHost()
+	assert.Nil(t, err)
+	loader := schema.NewPluginLoader(host)
+	defer contract.IgnoreClose(host)
+
+	pkgLoader := pulumiyaml.NewPackageLoaderFromSchemaLoader(loader)
+	_, diags, err = pulumiyaml.PrepareTemplate(template, nil, pkgLoader)
+	assert.Nil(t, err)
+	assert.Equal(t, "<stdin>:10,21-40: resource, variable, or config value \"bucketNamePrefix\" not found; ", diags.Error())
```
iwahbe, line 51: I don't understand. It looks like `bucketNamePrefix` exists here.

aq17: `ParseTemplate` alone should result in an error (that is swallowed), but yes it wouldn’t actually be called independently. I’ll remove this.

PR #409, opened 2022-11-17
Fix commit history

aq17:
https://github.com/pulumi/pulumi-yaml/pull/407 was merged without its commits squashed. This PR squashes those commits.


Frassle:
> This PR squashes those commits.

That won't help with history. The old non-squashed commits will still be there. If we actually _need_ to fix this someone with admin perms is going to need to turn off branch protection force push new history to main and then turn branch protection back on.
That's quite disruptive, so it might be we just live with these commits as is and try not to do this again (isn't there some checks from pu/pu workflows we can copy to here to lint that history is squashed?)

iwahbe:
I don't think we ever had an auto-squash in pulumi/pulumi. We just hit the squash-and-merge button instead of the merge button.

Frassle:
I think @AaronFriel added a check to PRs because bors can't do squash and merge so we warn if it's not pre-squashed (or maybe just not linear history...)

iwahbe:
We only warn if not linear history.

PR #411, opened 2022-11-18
Update examples to use `config`

### LINKED ISSUE #404: Update examples to use `config` instead of `configuration`
`pulumi convert` requires a bit more refactoring before `config` can be safely used and `configuration` can be fully deprecated.

aq17:
Fixes #404.


iwahbe:
It looks like we are getting a test failure here.

aq17:
We can probably loosen the constraint in `pu/pu` as well to accept `int`/`integer`.

iwahbe (approved):
LGTM when tests go green

PR #413, opened 2022-11-28
Loosen constraint on int vs number type check

aq17:
Fixes `pu/templates` https://github.com/pulumi/templates/issues/465

Since the project-level `config` block supports `integer` but not `number`, we can loosen the type check constraint for now until `configuration` is fully deprecated

### DIFF
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -1085,7 +1085,7 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	}
 	// check for incompatible types between config/ configuration
 	if typExisting, ok := tc.configuration[k]; ok {
-		if typExisting.String() != typ.String() {
+		if !tc.compatibleType(typExisting, typ) {
 			ctx := r.newContext(node)
 			ctx.error(nil, fmt.Sprintf("config key %s cannot have conflicting types %v, %v",
 				k, codegen.UnwrapType(typExisting), codegen.UnwrapType(typ)))
@@ -1094,6 +1094,15 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	}
 	tc.configuration[k] = typ
 	return true
+}
+
+func (tc *typeCache) compatibleType(a, b schema.Type) bool {
+	if a.String() == b.String() {
+		return true
+	} else if (a == schema.IntType && b == schema.NumberType) || (b == schema.IntType && a == schema.NumberType) {
+		return true
+	}
+	return false
 }
 
 func (tc *typeCache) typeOutput(r *Runner, node ast.PropertyMapEntry) bool {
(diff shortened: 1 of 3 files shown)


iwahbe:
Can you show a repro here?

Pulumi YAML `configuration` does support `int`:

https://github.com/pulumi/pulumi-yaml/blob/55d4ae7e7e57a952b4dc109f6a7ca9dfce86b430/pkg/pulumiyaml/config/config.go#L142-L144

pkg/pulumiyaml/analyser.go
```diff
@@ -1096,6 +1096,15 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	return true
 }
 
+func (tc *typeCache) compatibleType(a, b schema.Type) bool {
```
iwahbe, line 1099: I would rename this method to `isConfigEquivalent` or something similar. Its not clear without context what "compatibleType" means.

aq17:
> Can you show a repro here?
> 
> Pulumi YAML `configuration` does support `int`:
> 
> https://github.com/pulumi/pulumi-yaml/blob/55d4ae7e7e57a952b4dc109f6a7ca9dfce86b430/pkg/pulumiyaml/config/config.go#L142-L144

Ah right but `config` doesn't support `Number`. [The template before the change ](https://github.com/pulumi/templates/pull/466/files#diff-e09baca8af60415fbede053c04e35808b19090ff2b699754f07369640e7a9900) used `configuration` with `number`, and I assumed if `config` was set on the cli it was parsed as `integer` causing the conflict. I'll try to repro to confirm first

iwahbe (commented):
Not all numbers are integers, so I don't love this fix. I think a repro would be helpful in root causing this.

pkg/pulumiyaml/analyser.go
aq17: Is there a downside to defaulting integer -> number, or do we want to keep it strict?

iwahbe: There is a downside (we could allow programs that the user thinks we will reject), but it's minimal. I support allowing `integer -> number`, but not the other way around.

aq17:
The `Pulumi.dev.yaml` that gets initialized on `pulumi new container-aws-yaml` uses `config`, setting the keys as `integer`, whereas the old template with `configuration` tried to set it as `number`. Loosening the constraint to allow `integer` -> `number` should suffice for now

pkg/pulumiyaml/analyser.go
```diff
@@ -1065,37 +1065,49 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
...
+		} else {
+			typCurrent = typMatch
 		}
 	}
-	tc.configuration[k] = typ
+	tc.configuration[k] = typCurrent
 	return true
 }
 
+func (tc *typeCache) isConfigEquivalent(a, b schema.Type) (schema.Type, bool) {
```
iwahbe, line 1101: This method is no longer a test for type equivalence. Generally `is*` methods return `bool`. I read this as "unifyConfigType(a, b) -> (type, ok)".

If it doesn't depend on `tc`, we don't need it to be a method.

iwahbe (commented):
Would it be possible to add some tests here, both for correctness and to act as documentation for how the type coercion is supposed to work (int&float = float, int&int = int, int&string=!ok, ...)?

iwahbe (approved):

PR #416, opened 2022-12-01
Loosen constraint on config type checking

### LINKED ISSUE #415: Pulumi config value automatically interpreted as Integer
### What happened?

The config value get automatically set as integer from string. 

Pulumi version: v3.47.1

Input from CLI:

pulumi --non-interactive config set sha "123"

In the Pulumi.yaml:

configuration:
  sha:
    type: String

Error:
pulumi:pulumi:Stack testing-testing running Error: config key sha cannot have conflicting types string, integer

### Steps to reproduce

The config value get automatically set as integer from string. 

Pulumi version: v3.48.0 

Input from CLI:

pulumi --non-interactive config set sha "123"

In the Pulumi.yaml:

configuration:
  sha:
    type: String

Error:
pulumi:pulumi:Stack testing-testing running Error: config key sha cannot have conflicting types string, integer

### Expected Behavior

No config type error returned

### Actual Behavior

Error:
pulumi:pulumi:Stack testing-testing running Error: config key sha cannot have conflicting types string, integer

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

aq17:
Fixes #415 
Extension of https://github.com/pulumi/pulumi-yaml/pull/413

### DIFF
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -1055,7 +1055,8 @@ func (tc *typeCache) typeVariable(r *Runner, node variableNode) bool {
 }
 
 func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
-	k := node.key().Value
+	k, v := node.key().Value, node.value()
+	isConfigEnv := false
 	var typCurrent schema.Type = &schema.InvalidType{}
 	var optional bool
 
@@ -1077,6 +1078,7 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 		if n.Type != nil {
 			typCurrent = n.Type.Schema()
 		}
+		isConfigEnv = true
 	}
 
 	typCurrent = &schema.InputType{ElementType: typCurrent}
@@ -1085,9 +1087,13 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	}
 	// check for incompatible types between config/ configuration
 	if typExisting, ok := tc.configuration[k]; ok {
-		if typMatch, ok := unifyConfigType(typExisting, typCurrent); !ok {
+		var curVal interface{}
+		if isConfigEnv {
+			curVal = v
+		}
+		if typMatch, ok := unifyConfigType(typExisting, typCurrent, curVal); !ok {
 			ctx := r.newContext(node)
-			ctx.error(nil, fmt.Sprintf("config key %s cannot have conflicting types %v, %v",
+			ctx.error(nil, fmt.Sprintf(`config key "%s" cannot have conflicting types %v, %v`,
 				k, codegen.UnwrapType(typExisting), codegen.UnwrapType(typCurrent)))
 			return false
 		} else {
@@ -1098,13 +1104,32 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	return true
 }
 
-// checks for config type compatibility
-func unifyConfigType(a, b schema.Type) (schema.Type, bool) {
-	a, b = codegen.UnwrapType(a), codegen.UnwrapType(b)
-	if a.String() == b.String() {
-		return a, true
-	} else if (a == schema.IntType && b == schema.NumberType) || (b == schema.IntType && a == schema.NumberType) {
+// Checks for config type compatibility between an already registered/ old config key with a new key.
+// We allow Int to fall back to Number.
+// We also allow a new non-String key to fall back to String
+// since parsing CLI-set config types is a best effort and may actually be intended as a String.
+// If the new key is a String with a defined value, and the old key is not, then we
+// parse the new key to determine if it can fall back to the old key's type
+// (i.e. "16" would be valid as an Int but "foo" would not.)
+// TODO: remove the last case once `configuration` is deprecated.
+func unifyConfigType(oldType, newType schema.Type, newVal interface{}) (schema.Type, bool) {
+	oldType, newType = codegen.UnwrapType(oldType), codegen.UnwrapType(newType)
+	if oldType.String() == newType.String() {
+		return oldType, true
+	} else if (oldType == schema.IntType && newType == schema.NumberType) || (newType == schema.IntType && oldType == schema.NumberType) {
 		return schema.NumberType, true
+	} else if oldType == schema.StringType {
+		return schema.StringType, true
+	} else if newType == schema.StringType {
+		if v, ok := newVal.(string); !ok {
+			return nil, false
+		} else if _, err := strconv.ParseInt(v, 10, 64); err == nil && oldType == schema.IntType {
+			return schema.IntType, true
+		} else if _, err := strconv.ParseBool(v); err == nil && oldType == schema.BoolType {
+			return schema.BoolType, true
+		} else if _, err := strconv.ParseFloat(v, 64); err == nil && oldType == schema.NumberType {
+			return schema.NumberType, true
+		}
 	}
 	return nil, false
 }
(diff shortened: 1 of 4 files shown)


aq17:
This is definitely not the most elegant fix but think it should suffice until `configuration` is deprecated

iwahbe (changes requested):
I haven't tested it, but I'm pretty sure this will lead to undesirable behavior. It will solve the type unification error by coercing the `configuration` type to a `schema.StringType`, which will then error as the wrong type if it is used as any other type.

I think an example will help illustrate:

```yaml
config:
  myStack:length: "16" # Of type string
configuration:
  length:
    type: int # Coerced to type string in `unifyConfigType`
resources:
  password:
    type: random:RandomPassword
    properties:
      length: ${length} # Type error because `${length}` is of type string
                        # Especially bad because the error contradicts the type declaration right above
```

I think that `unifyConfigType` needs to take values into consideration.

"16" & number => number, because we can confirm that "16" can be a number.
"foo" & number => error, because we know that there will be an error down the line.

aq17:
> I haven't tested it, but I'm pretty sure this will lead to undesirable behavior. It will solve the type unification error by coercing the `configuration` type to a `schema.StringType`, which will then error as the wrong type if it is used as any other type.
> 
> I think an example will help illustrate:
> 
> ```yaml
> config:
>   myStack:length: "16" # Of type string
> configuration:
>   length:
>     type: int # Coerced to type string in `unifyConfigType`
> resources:
>   password:
>     type: random:RandomPassword
>     properties:
>       length: ${length} # Type error because `${length}` is of type string
>                         # Especially bad because the error contradicts the type declaration right above
> ```
> 
> I think that `unifyConfigType` needs to take values into consideration.
> 
> "16" & number => number, because we can confirm that "16" can be a number. "foo" & number => error, because we know that there will be an error down the line.

Yes I initially thought about comparing values alongside types but wasn't sure how to do so for `configuration` keys – on second thought, I think we only need to keep track of `config`. values.
Since `configuration` is guaranteed to have either an explicit default or type, if "16" is declared by the CLI and interpreted as an int, but is defined as a string by `configuration`, then the int->number is a safe assumption (current implementation).
The issue only arises in the reverse case (such as your example), in which case we need to check that `config` has "16" and not "foo".

Does this sound reasonable?

iwahbe:
> > I haven't tested it, but I'm pretty sure this will lead to undesirable behavior. It will solve the type unification error by coercing the `configuration` type to a `schema.StringType`, which will then error as the wrong type if it is used as any other type.
> > I think an example will help illustrate:
> > ```yaml
> > config:
> >   myStack:length: "16" # Of type string
> > configuration:
> >   length:
> >     type: int # Coerced to type string in `unifyConfigType`
> > resources:
> >   password:
> >     type: random:RandomPassword
> >     properties:
> >       length: ${length} # Type error because `${length}` is of type string
> >                         # Especially bad because the error contradicts the type declaration right above
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that `unifyConfigType` needs to take values into consideration.
> > "16" & number => number, because we can confirm that "16" can be a number. "foo" & number => error, because we know that there will be an error down the line.
> 
> Yes I initially thought about comparing values alongside types but wasn't sure how to do so for `configuration` keys – on second thought, I think we only need to keep track of `config`. values. Since `configuration` is guaranteed to have either an explicit default or type, if "16" is declared by the CLI and interpreted as an int, but is defined as a string by `configuration`, then the int->number is a safe assumption (current implementation). The issue only arises in the reverse case (such as your example), in which case we need to check that `config` has "16" and not "foo".
> 
> Does this sound reasonable?

I think we can do a one way conversion, where we coerce `type(config[var])` to `type(configuration[var])` but not vice versa. Since we always have the value for `config[var]`, this is easier.

pkg/pulumiyaml/analyser.go
```diff
@@ -1085,9 +1087,13 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
 	}
 	// check for incompatible types between config/ configuration
 	if typExisting, ok := tc.configuration[k]; ok {
-		if typMatch, ok := unifyConfigType(typExisting, typCurrent); !ok {
+		var curVal interface{}
+		if isConfigEnv {
+			curVal = v
+		}
+		if typMatch, ok := unifyConfigType(typExisting, typCurrent, curVal); !ok {
```
iwahbe, line 1094: Is there a reason not to pass `node.value()` in all cases here?
```suggestion
		if typMatch, ok := unifyConfigType(typExisting, typCurrent, node.value()); !ok {
```

pkg/pulumiyaml/analyser.go
```diff
@@ -1098,13 +1104,32 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
...
-		return a, true
-	} else if (a == schema.IntType && b == schema.NumberType) || (b == schema.IntType && a == schema.NumberType) {
+// Checks for config type compatibility between an already registered/ old config key with a new key.
+// We allow Int to fall back to Number.
+// We also allow a new non-String key to fall back to String
+// since parsing CLI-set config types is a best effort and may actually be intended as a String.
+// If the new key is a String with a defined value, and the old key is not, then we
+// parse the new key to determine if it can fall back to the old key's type
+// (i.e. "16" would be valid as an Int but "foo" would not.)
+// TODO: remove the last case once `configuration` is deprecated.
```
iwahbe, line 1114: I think this reads clearer, but you are free to ignore me here.

```suggestion
// Config types are compatible if
// - They are the same type.
// - We are assigning an integer to a number.
// - We are assigning any type to a string.
// - We are assigning a string to some type T *and* the string can be unambiguously parsed into T.
// TODO: remove the last case once `configuration` is deprecated.
```

pkg/pulumiyaml/analyser.go
```diff
@@ -1098,13 +1104,32 @@ func (tc *typeCache) typeConfig(r *Runner, node configNode) bool {
...
+func unifyConfigType(oldType, newType schema.Type, newVal interface{}) (schema.Type, bool) {
+	oldType, newType = codegen.UnwrapType(oldType), codegen.UnwrapType(newType)
+	if oldType.String() == newType.String() {
+		return oldType, true
+	} else if (oldType == schema.IntType && newType == schema.NumberType) || (newType == schema.IntType && oldType == schema.NumberType) {
```
iwahbe, line 1119: Since `unifyConfigType` is now directional, we can constrain this conversion to be one way. I'm pretty sure this will handle the use case we want: `config number` to `configuration int`.

```suggestion
	} else if newType == schema.IntType && oldType == schema.NumberType {
```


If you agree with the above change, then `unfiyConfigType(oldType, newType, newValue)` will always return `oldType, true` or `nil, false` and we can reduce its signature to `bool` (with an associated name change).

pkg/pulumiyaml/analyser.go
aq17: `node.value()` for a `configuration` key would be a `*ast.ConfigParamDecl`, but I realize now that's irrelevant. Also I believe we only hit conflicts where we are registering a `config` key but already have a matching `configuration` key since we store in-memory `configuration` then `config` keys`.

pkg/pulumiyaml/analyser.go
aq17: LGTM. Although "assigning" might be a bit ambiguous, but I think the test cases help clarify it as well.

pkg/pulumiyaml/analyser.go
aq17: 👍🏼 Again I think it relies on the assumption that conflicts strictly arise when attempting to assign a `config` key to an already-registered `configuration` key but that should be safe

iwahbe (approved):
I left a couple nits. After that, LGTM!

PR #419, opened 2022-12-05
Deprecate fn::stackReference

### LINKED ISSUE #285: Deprecate Fn::StackReference
Ahead of a 1.0 launch we should:

- [x] (Optional?) Warn on `Fn::StackReference` 
- [ ] Error on `Fn::StackReference`

In either case it'll be simpler to point to the docs than emit an error message with a syntax fix.

aq17:
Fixes #285 
Now that YAML is GA, `fn::stackReference` should be deprecated in favor of using the resource type `pulumi:pulumi:StackResource`.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,5 @@
 ### Improvements
+- Deprecate `fn::stackReference`.
+  [#419](https://github.com/pulumi/pulumi-yaml/pull/419)
 
 ### Bug Fixes
--- a/examples/stackreference-consumer/Pulumi.yaml
+++ b/examples/stackreference-consumer/Pulumi.yaml
@@ -13,4 +13,4 @@ resources:
       # So you might use "alice/infra-project-name/dev", as an example.
       name: PLACEHOLDER_ORG_NAME/stackreference-producer/PLACEHOLDER_STACK_NAME
 outputs:
-  referencedImageName: ${stackRef.imageName}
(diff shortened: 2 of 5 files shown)


iwahbe:
I think we need to go through a depreciation step here before we remove stack references. I would update stack reference example and add a warning that links to (a cleaned up) #285 when the old syntax is used. If we remove a feature, we need to warn users before things stop working. Right now, we don't warn when `fn::stackReference` is used.

aq17:
> I think we need to go through a depreciation step here before we remove stack references. I would update stack reference example and add a warning that links to (a cleaned up) #285 when the old syntax is used. If we remove a feature, we need to warn users before things stop working. Right now, we don't warn when `fn::stackReference` is used.

Agreed –  I thought we had did have a warning previously before updating all of our docs to the stack reference resource type. If not, what should our timeline be between warning and deprecation?

iwahbe:
> > I think we need to go through a depreciation step here before we remove stack references. I would update stack reference example and add a warning that links to (a cleaned up) #285 when the old syntax is used. If we remove a feature, we need to warn users before things stop working. Right now, we don't warn when `fn::stackReference` is used.
> 
> Agreed – I thought we had did have a warning previously before updating all of our docs to the stack reference resource type. If not, what should our timeline be between warning and deprecation?

I thought we did too, but testing `fn::stackReference` didn't have one. It's generally best to confirm yourself. For the depreciation timeline, I'll defer to @AaronFriel.

PR #420, opened 2022-12-06
Deprecate `fn::stackReference`

### LINKED ISSUE #285: Deprecate Fn::StackReference

aq17:
Fixes #285 

Now that YAML is GA, `fn::stackReference` should be deprecated in favor of using the resource type pulumi:pulumi:StackResource.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,5 @@
 ### Improvements
+- Deprecate `fn::stackReference`.
+  [#420](https://github.com/pulumi/pulumi-yaml/pull/420)
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,5 @@
 ### Improvements
+- Deprecate `fn::stackReference`.
```
iwahbe, line 2: I think we normally have a space here.
```suggestion

- Deprecate `fn::stackReference`.
```

iwahbe (approved):
LGTM!

PR #429, opened 2022-12-20
Update README examples

Zaid-Ajaj:
- `bucket: ${my-bucket}` -> `bucket: ${my-bucket.id}` so that `pulumi convert` works with csharp target from the get-go
- `protect: true` to be nested under `options`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ resources:
   index.html:
     type: aws:s3:BucketObject
     properties:
-      bucket: ${my-bucket}
+      bucket: ${my-bucket.id}
       source:
         fn::stringAsset: <h1>Hello, world!</h1>
       acl: public-read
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ resources:
   index.html:
     type: aws:s3:BucketObject
     properties:
-      bucket: ${my-bucket}
+      bucket: ${my-bucket.id}
```
AaronFriel, line 23: Ensuring this behavior works I think is important to the ergonomics of Pulumi YAML.

README.md
```diff
@@ -58,7 +58,8 @@ resources:
           fromPort: 80
           toPort: 80
           cidrBlocks: ["0.0.0.0/0"]
-    protect: true
+    options:
+      protect: true
```
AaronFriel, line 62: Good change 👍

AaronFriel (changes requested):
I think I'd rather change the YAML to PCL intermediate than the example, as this is an intentional feature of the language.

PR #431, opened 2022-12-21
Introduce `fn::method` function for resource methods

### LINKED ISSUE #354: Support calling resource methods
It is almost possible to call resource methods manually, as in this program which calls the "getKubeconfig" method on cluster:

```yaml
variables:
  kubeconfig:
    Fn::Invoke:
      Function: google-native:container/v1:Cluster/getKubeconfig
      Arguments:
        __self__: ${someClusterResource}
```

However, this generates an `Invoke`, not a `Call` RPC. The syntax also leaves something to be desired, with the `__self__` parameter being an implementation detail that other Pulumi languages do not expose to users.

aq17:
Fixes #354 
Allows resource methods to be called via `fn::method`, using the following syntax:
```
variables:
  kubeconfig:
    fn::method:
      resource: ${someClusterResource}
      function: getKubeconfig
      arguments:
        ...
```
Both the full function type token or just the function name may be used


iwahbe:
It would be great if you could write up which syntax you have implemented, as there were several suggestions on the original issue. A test would likewise make the desired behavior clearer.

One thing comes to mind immediately. We have `CallOpts` of type `InvokeOptionsDecl`. I think this means we are allowing users to invoke a method on a resource of a different version/provider. I'm not sure this makes sense. @justinvp would know more if this should be allowed.

justinvp:
> One thing comes to mind immediately. We have `CallOpts` of type `InvokeOptionsDecl`. I think this means we are allowing users to invoke a method on a resource of a different version/provider. I'm not sure this makes sense. @justinvp would know more if this should be allowed.

Yeah, I wouldn't expect we'd allow users to pass these. It should be passed along from the resource itself.

aq17:
@AaronFriel update: `pulumi up` should be working with the syntax we discussed with @justinvp
Todo: 
- testing: can we extend [run_invoke_test.go ](https://github.com/pulumi/pulumi-yaml/blob/main/pkg/pulumiyaml/run_invoke_test.go#L218), or do we need a separate framework?
- codegen/ PCL representation

pkg/pulumiyaml/run.go
```diff
@@ -1758,6 +1761,48 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
...
+						functionName = FunctionTypeToken(m.Function.Token)
+						matched = true
+						break
+					}
+				}
+			}
+			if !matched {
+				return e.error(t, err.Error())
+			}
+		}
```
iwahbe, line 1794: What is this doing? It looks like we are doing type-checking within the apply body.

A comment explaining what is going on here would be appreciated.

iwahbe (commented):
Resource methods will need to be type checked, along with their arguments.

AaronFriel:
@aq17 unfortunately that will require extending the `MockResourceMonitor`:

https://github.com/pulumi/pulumi/blob/9d7ccc628b2fed47ac0da67b1526bb2fed265acc/sdk/go/pulumi/mocks.go#L113-L158

It looks like due to a fluke or naming mistake, that interface uses a method named `Call` and implementations typically implement a function field named `CallF` to implement the `Invoke` RPC. That'll make adding support for mocking the `Call` RPC a little unusual. 

@justinvp what do you think?

justinvp:
> @justinvp what do you think?

This is https://github.com/pulumi/pulumi/issues/8328

(Yes, it is unfortunate that the mock name for invokes is `Call` rather than `Invoke`.) If we were to add support for mocking the gRPC `Call`, I think the simplest thing to do would be to make the gRPC `Call` call the mock `Call`. It might seem a little strange, but having the `Invoke` and `Call` gRPCs both call the mock `Call` method is similar to how the `RegisterResource` and `ReadResource` gRPCs both call the mock `NewResource` method. (And if we needed to, at some point, we could add more fields to `MockCallArgs` to indicate whether it's an `Invoke` or `Call` gRPC, etc.)

The alternative is doing more design work to have a separate method in the mock just for `Call`. Not sure what we'd name it, though.

pkg/pulumiyaml/run.go
aq17: I think I've moved the type-checking to the proper place, storing a lookup table for resource methods to validate a resource+method

t0yv0:
👋 any updates on this, bumping into this in the context of https://github.com/t0yv0/pulumi-12709 which special cases for single-value Resource returning methods, and YAML is part of the requirement. CC @lblackstone 

It sounds like there's no current workaround to call the methods in YAML, it just is not possible. Considering options here, any potential workarounds or perhaps would it be feasible for me to pick up this PR and bring it to completion. I haven't contributed to this codebase yet but can get ramped up if needed. That's an option..

I can quickly look at the possibility if YAML can pass an `Output<Resource>` that is a component resource output, not method call, to another resource's `provider:` option. This would be asymmetric to other languages but could be acceptable.

iwahbe:
Getting this in would make testing `Call` much easier.

PR #436, opened 2023-01-24
Re-enable go linter

iwahbe:
Re-enable the linter.


AaronFriel (approved):

.github/workflows/stage-lint.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Lint
         run: |
           curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.49.0
-          make lint-golang || true
+          make lint-golang
```
iwahbe, line 32: We were running the linter, then throwing away the result.

Frassle:
We're running golangci-lint on main now so I think this is done.

PR #437, opened 2023-01-28
Add support for the `deletedWith` resource option

### LINKED ISSUE #432: Add support for the `DeletedWith` resource option
## Hello!

## Issue details

https://github.com/pulumi/pulumi/pull/11095 added support for a new `DeletedWith` resource option to the Go, Node.js, Python, and .NET SDKs. This option should be available from Java as well.

justinvp:
This adds support for the `deletedWith` resource option that has recently been added to the Node.js, Python, .NET, and Go SDKs.

In many cases there is no need to deleted resources if the container resource is going to be deleted as well.

A few examples:
- Database object (roles, tables) when database is being deleted
- Cloud IAM bindings when user itself is being deleted

This helps with:
- Speeding the deletion process
- Removing unnecessary calls to providers
- Avoiding failed deletions when the pulumi user running the plan has access to the container resource but not the contained ones

To avoid deleting contained resources, set the `deletedWith` resource option to the container resource.

Fixes #432


iwahbe (approved):
This looks like a textbook change. LGTM.

P.S. I would look into the upgraded version of pu/pu when investigating why the tests fail.

aq17:
@justinvp I added a small commit and associated PR in [pu/pu](https://github.com/pulumi/pulumi/pull/12011) that should allow the option to be passed on `pulumi convert`

pkg/pulumiyaml/codegen/load.go
```diff
@@ -857,6 +857,13 @@ func (imp *importer) importResource(kvp ast.ResourcesMapEntry, latestPkgInfo map
 		}
 	}
 
+	if resource.Options.DeletedWith != nil {
+		resourceOptions.Body.Items = append(resourceOptions.Body.Items, &model.Attribute{
```
justinvp, line 861: @aq17, since `DeletedWith` takes a Resource, shouldn't this look more like the `if resource.Options.Parent != nil` case above? Or am I misunderstanding how this should work?

```go
	if resource.Options.DeletedWith != nil {
		ref, err := imp.getResourceRefItem(resource.Options.DeletedWith, name, "deletedWith")
		if err != nil {
			diags.Extend(err)
		} else {
			resourceOptions.Body.Items = append(resourceOptions.Body.Items, &model.Attribute{
				Name:  "deletedWith",
				Value: ref,
			})
		}
	}
```

aq17: Yup this was incorrectly based on `DeletedWith` still accepting the URN as a string

PR #451, opened 2023-04-25
Remove S3 public access block in examples

AaronFriel:
This unblocks these examples working as simple static sites. The changes in this commit echo those made in our AWS getting started guide here:
- pulumi/pulumi-hugo#2720

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,19 @@ resources:
     properties:
       website:
         indexDocument: index.html
+  # The ownershipControls and publicAccessBlock resources are required as of April 2023
+  # https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/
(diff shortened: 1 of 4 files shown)


cnunciato (approved):

AaronFriel:
@cnunciato It looks like the BucketPolicy got an AccessDenied, I saw that locally but it went away on 2nd push. Do I need to add a DependsOn there, too?

Or do we need this policy ACL at all for the example?

```yaml

  bucketPolicy:
    type: aws:s3:BucketPolicy
    properties:
      bucket: ${site-bucket}
      policy: |
        {
          "Version": "2012-10-17",
          "Statement": [
            {
              "Effect": "Allow",
              "Principal": "*",
              "Action": ["s3:GetObject"],
              "Resource": ["${site-bucket.arn}/*"]
            }
          ]
        }
```

PR #471, opened 2023-08-01
go.mod: Tidy with Go 1.19

abhinav:
Runs `go mod tidy` with -go=1.19 on the go.mod for the project.
Note that as of Go 1.20, the `go` directive is **not**
the minimum required Go version.
It's the version of the toolchain we developed against.
This is changing in Go 1.21.

The `-compat` flag specifies which version of Go the go.mod file
needs to be compatible with.
By default, this is one release behind the `-go` flag
(or `go` directive).
So we don't need to specify it manually.

Lastly, this deletes the `go mod tidy` in the test job.
Only the lint job should bbe modifying the go.mod
to ensure that the go.mod is up to date.

### DIFF
--- a/.github/workflows/stage-lint.yml
+++ b/.github/workflows/stage-lint.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: 1.20.x
-      - run: go mod tidy -compat=1.17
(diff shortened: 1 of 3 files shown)


abhinav:
Oops. This was not intended to be merged into the other branch.

PR #500, opened 2023-08-27
Add pulumi-converter-yaml

Frassle:
The CLI won't actually use this yet because it's hardcoded to call into the directly linked Eject function for "language=yaml", but getting this added and published is the first step towards getting the CLI to use it instead of depending on the yaml code directly.


cmd/pulumi-converter-yaml/main.go
```diff
@@ -0,0 +1,115 @@
...
+// responsible for translating JSON/YAML templates into PCL.
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"path/filepath"
+
+	"github.com/pkg/errors"
```
abhinav, line 25: Was this use of pkg/errors intentional?

Frassle: no, probably vscode being overly eager to auto-fill. I'll fix it up.

cmd/pulumi-converter-yaml/main.go
```diff
@@ -0,0 +1,115 @@
...
+func (*yamlConverter) ConvertState(ctx context.Context,
+	req *plugin.ConvertStateRequest,
+) (*plugin.ConvertStateResponse, error) {
+	return nil, errors.New("not implemented")
+}
+
+// writeProgram writes a project and pcl program to the given directory
+func writeProgram(directory string, proj *workspace.Project, program *pcl.Program) error {
+	fs := afero.NewOsFs()
+	err := program.WriteSource(afero.NewBasePathFs(fs, directory))
```
abhinav, line 54: minor: This function should take the file system as an argument instead of building its own.
It can probably take just the fs, not the directory; letting the caller handle creating the base path FS.

```go
func writeProgram(fs afero.Fs, proj *workspace.Project, program *pcl.Program) error
```

cmd/pulumi-converter-yaml/main.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return fmt.Errorf("writing program: %w", err)
+	}
+
+	// Write out the Pulumi.yaml file if we've got one
+	if proj != nil {
```
abhinav, line 60: nit: I would invert the if condition here; proj being nil is  a special case so we can handle that early.

```go
if proj == nil { return nil }
```

And then the rest of the logic below gets unnested.

cmd/pulumi-converter-yaml/main.go
```diff
@@ -0,0 +1,115 @@
...
+		projBytes, err := encoding.YAML.Marshal(proj)
+		if err != nil {
+			return fmt.Errorf("marshaling project: %w", err)
+		}
+
+		err = afero.WriteFile(fs, filepath.Join(directory, "Pulumi.yaml"), projBytes, 0o644)
```
abhinav, line 66: If the function accepts just the fs, we won't need to do this join since the fs is already rooted in the directory that we care about.

If you choose not to take the fs an argument, and keep the directory, I would still recommend at least this:

```go
fs := afero.NewBasePathFs(afero.NewOsFs(), directory)
```

So that the rest of the function never joins on directory again.

abhinav (approved):
No blockers, just minor improvements.

PR #509, opened 2023-10-10
respect types for config set in the yaml file

### LINKED ISSUE #453: Quoted variable values incorrectly converted to integers
### What happened?

My Pulumi.develop.yaml file

```yaml
config:
  aws:profile: 028111111111_AWSAdministratorAccess
  aws:region: ap-southeast-2
  myprojectname:aws_account_id: "028111111111" # I also tried single quotes here
  myprojectname:dns_domain: "redacted"
  myprojectname:aws_route53_zoneid: "redacted"
```

When I attempt to use this value `aws_account_id` in the code, it strips the zero. The stripping doesn't happen to the `aws:profile`. Perhaps because it contains some alpha characters too, or perhaps because the aws provider handles the variable differently.

Someone on Slack (Scott Lowe) suggested:

> Do you have a reference to that configuration value in the config block of your `Pulumi.yaml` file? If so, you can specify a type property there (set it to “string”), and that may help resolve the issue.

I tried that, perhaps I'm missing something simple.
So this value` aws_account_id` is different for each stack hence its original location in the stack files i.e. `Pulumi.develop.yaml`, `Pulumi.staging.yaml` etc.
I tried to implement your suggestion by adding a couple of lines under config: at the top of `Pulumi.yaml` like so:
config:
```yaml
  aws_account_id:
    type: string
```
Then I printed the output like so:

```yaml
outputs:
  # Debug printouts
  accIDString: "${aws_account_id}"
```

And it's still getting converted.

```log
Outputs:
  + accIDString               : 2.809102044e+10
```

I also tried a slightly different way, but this just printed the variable name

```yaml
config:
  awsAccountID: 
    type: string
    default: ${aws_account_id}
...
outputs:
  # Debug printouts
  accIDString: "${awsAccountID}"
~~~~~~~~
Outputs:
  + accIDString               : "${aws_account_id}"
```

### Expected Behavior

I expected variable values entered with quotes to stay as strings

### Steps to reproduce

Try to use the AWS account value in the yaml code

This is where I tried to use it (in an IAM policy)

```yaml
 # Allow to send emails with SES.
            - Action:
                - ses:SendEmail
              Effect: Allow
              Resource: "*" # "arn:aws:ses:${aws:region}:${aws_account_id}:identity/${ses_DomainIdentity.domain}" # can't use this due to bug with yaml stripping leading 0 from aws_account_id
```

### Output of `pulumi about`

```
CLI          
Version      3.65.1
Go Version   go1.20.3
Go Compiler  gc

Plugins
NAME  VERSION
yaml  unknown

Host     
OS       debian
Version  11.5
Arch     x86_64

This project is written in yaml
```

### Additional context

_No response_

### Contributing

tgummerer:
The yaml language provider currently gets the config via RPC and then
using some magic to try and infer the type of the variable it gets as
a string over the RPC interface.  However in many cases that's not
quite what the user wants, e.g. someone might have an AWS account ID
that starts with a `0`.  If that gets interpreted as a number the
account ID will no longer be correct.

Prefer the config nodes we parse out of the `Pulumi.yaml` file over
the ones we get through the RPC, falling back to the RPC ones when we
don't have more information in the yaml.

This feels a little bit awkward still, and maybe we should be passing the types through the RPC?  Curious to hear peoples opinions about this.

Fixes #453


Frassle:
> This feels a little bit awkward still, and maybe we should be passing the types through the RPC? Curious to hear peoples opinions about this.

We probably should. All the other languages have to deal with _just_ the RPC protocol, they don't (and shouldn't) look at the config yaml files. 
Plus with esc and project config the RPC protocol has _far_ better information than the yaml plugin can hope to keep up with.

tgummerer:
I think this is ready for another review, now taking the types from the RPC call.  Before merging this it needs https://github.com/pulumi/pulumi/pull/14244, which is also required to get the tests pass here.

I'm not sure here, do we usually wait for a new release of the CLI so the tests here can pass with that?  Or would it be preferable to just make the action install the particular SHA?

Frassle:
> I'm not sure here, do we usually wait for a new release of the CLI so the tests here can pass with that? Or would it be preferable to just make the action install the particular SHA?

It's fine to depend on the sha _once_ it's merged to master. We'd need to release a tag for yaml and update pu/pu's dependency as well to actually get this fix shipped.

tgummerer:
Closing in favor of https://github.com/pulumi/pulumi-yaml/pull/511.

PR #510, opened 2023-10-10
Allow version to be set dynamically

### LINKED ISSUE #508: Unable to version an explicit provider from config
### What happened?

Trying to set the version of an explicit provider from outside the stack, running into unexpected error messages from Pulumi YAML. Is there a way to do this?

### Example

```
name: test

runtime:
  name: yaml

config:
  version:
    type: string

resources:

  provider:
    type: pulumi:providers:aws
    options:
      version: ${version}

  tested-resource:
    type: aws:s3/bucket:Bucket
    properties:
      bucket: testbucket-${pulumi.stack}
    options:
      provider: ${provider}
```

```
pulumi config set version 5.42.0
pulumi preview
```

```
error: failed to discover plugin requirements: Pulumi.yaml:15,16-26: version must be a string; 
```

### Output of `pulumi about`

```
CLI          
Version      3.86.0
Go Version   go1.21.1
Go Compiler  gc

Host     
OS       darwin
Version  14.0
Arch     x86_64

This project is written in yaml

Current Stack: t0yv0/test/buckett

Found no resources associated with buckett

Found no pending operations associated with buckett

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/t0yv0
User           t0yv0
Organizations  t0yv0, pulumi
Token type     personal

No dependencies found

Pulumi locates its logs in /var/folders/gk/cchgxh512m72f_dmkcc3d09h0000gp/T/ by default
```

### Additional context

_No response_

### Contributing

Zaid-Ajaj:
Fixes pulumi/pulumi-yaml#508

This PR changes the definition of `Version` from a static `*ast.StringExpr` to `ast.Expr` so that it can a value which can be evaluated at runtime when running the program rather than just static strings. Anywhere else where want to know the version and we don't have an evaluator yet, we check if the version is a static string and get its value, otherwise we skip it.

### DIFF
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/blang/semver"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
@@ -563,11 +564,27 @@ func (tc *typeCache) assertTypeAssignable(ctx *evalContext, from ast.Expr, to sc
 func (tc *typeCache) typeResource(r *Runner, node resourceNode) bool {
 	k, v := node.Key.Value, node.Value
 	ctx := r.newContext(node)
-	version, err := ParseVersion(v.Options.Version)
-	if err != nil {
-		ctx.error(v.Type, fmt.Sprintf("unable to parse resource %v provider version: %v", k, err))
-		return true
+
+	var version *semver.Version
+	if v.Options.Version != nil {
+		// compute the type of the version field
+		tc.typeExpr(ctx, v.Options.Version)
+
+		// make sure specified version is typed as a string
+		tc.assertTypeAssignable(ctx, v.Options.Version, schema.StringType)
+
+		// if it happens to be a static string, make sure it parsable as a semver
+		switch versionValue := v.Options.Version.(type) {
+		case *ast.StringExpr:
+			parsedVersion, err := semver.ParseTolerant(versionValue.Value)
+			if err != nil {
+				ctx.error(v.Type, fmt.Sprintf("unable to parse resource %v provider version: %v", k, err))
+				return true
+			}
+			version = &parsedVersion
+		}
 	}
+
 	pkg, typ, err := ResolveResource(ctx.pkgLoader, v.Type.Value, version)
 	if err != nil {
 		ctx.error(v.Type, fmt.Sprintf("error resolving type of resource %v: %v", k, err))
@@ -740,6 +757,11 @@ func (tc *typeCache) typePropertyEntries(ctx *evalContext, resourceName, resourc
 }
 
 func (tc *typeCache) typeInvoke(ctx *evalContext, t *ast.InvokeExpr) bool {
+	if t.CallOpts.Version != nil {
+		tc.typeExpr(ctx, t.CallOpts.Version)
+		tc.assertTypeAssignable(ctx, t.CallOpts.Version, schema.StringType)
+	}
+
 	version, err := ParseVersion(t.CallOpts.Version)
 	if err != nil {
 		ctx.error(t.CallOpts.Version, fmt.Sprintf("unable to parse function provider version: %v", err))
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -19,6 +19,8 @@ import (
 	"sync"
 	"unicode/utf8"
 
+	"github.com/blang/semver"
+
 	"github.com/google/shlex"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/hcl/v2"
@@ -197,7 +199,7 @@ func (r *Runner) validateResources() {
 // This function communicates errors by appending to the internal diags field of `r`.
 // It is the responsibility of the caller to verify that no err diags were appended if
 // that should prevent proceeding.
-func (r *Runner) setDefaultProviders() {
+func (r *Runner) setDefaultProviders(e programEvaluator) {
 	defaultProviderInfoMap := make(map[string]*providerInfo)
 	for _, resource := range r.t.Resources.Entries {
 		v := resource.Value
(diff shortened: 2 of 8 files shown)


pkg/pulumiyaml/run_options_test.go
```diff
@@ -64,9 +64,13 @@ configuration:
...
     default: false
     type: boolean
+  version:
+    default: "1.0.0"
+    type: string
 resources:
   provider-a:
     type: pulumi:providers:test
+    version: ${version}
   provider-b:
```
Frassle, line 74: Shouldn't something later in this test assert that provider-a is actually registered with version "1.0.0"?

pkg/pulumiyaml/semver.go
```diff
@@ -7,15 +7,20 @@ import (
 	"github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/ast"
 )
 
-func ParseVersion(v *ast.StringExpr) (*semver.Version, error) {
-	if v == nil || v.Value == "" {
-		return nil, nil
-	}
+func ParseVersion(expr ast.Expr) (*semver.Version, error) {
```
Frassle, line 10: Is this even used anymore? If we're allowing version to be any expression it probably isn't safe to have a function laying around that looks like it correctly gives a *semver.Version from an Expr.

Zaid-Ajaj: This is used in a few places where there isn't an evaluator available to evaluate the expression, so it tries to get the value of the version when it is statically known (i.e. `*ast.StringExpr`). 

>  it probably isn't safe to have a function laying around that looks like it correctly gives a *semver.Version from an Expr

The function isn't total, it will try to return a `*semver.Version` when it can from a string expression, otherwise it will return `nil` which the rest of the caller code handles already (when resolving resource/invoke versions) 

Maybe adding a `bool` to the returned tuple could make the signature a bit more clear. Returning `false` when the `Expr` is anything but `*ast.StringExpr`

iwahbe: Are we allowing `version` to be any expression, or just a non-literal? Specifically, are we allowing `version` to be a outputty value?

I would strongly recommend that `version` must be knowable at preview time.

pkg/pulumiyaml/run.go
```diff
@@ -1800,8 +1846,17 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
 	var opts []pulumi.InvokeOption
 
 	if t.CallOpts.Version != nil {
-		opts = append(opts, pulumi.Version(t.CallOpts.Version.Value))
+		if versionExpr, ok := e.evaluateExpr(t.CallOpts.Version); ok {
+			if version, ok := versionExpr.(string); ok {
+				opts = append(opts, pulumi.Version(version))
+			} else {
+				e.error(t.CallOpts.Version, "version must be a string value")
+			}
```
iwahbe, line 1854: This should be an internal error. It version resolves to something not a string, then the type checker should have caught it.

pkg/pulumiyaml/run.go
```diff
@@ -1230,6 +1255,27 @@ func (e *programEvaluator) registerResource(kvp resourceNode) (lateboundResource
...
+		if ok {
+			if !hasOutputs(versionValue) {
+				if version, ok := versionValue.(string); ok {
+					opts = append(opts, pulumi.Version(version))
+				} else {
+					e.error(v.Options.Version, "version must be a string value")
+					overallOk = false
+				}
+			} else {
+				e.error(v.Options.Version, "version must be not be an output")
```
iwahbe, line 1270: I don't remember if the type checker can catch this kind of thing, but if it can that would be great.

pkg/pulumiyaml/semver.go
Zaid-Ajaj: Syntactically it can be any expression but the evaluation should return `string` at runtime. We check that when we run `evaluateExpr`:
```go
if versionExpr, ok := e.evaluateExpr(t.CallOpts.Version); ok {
	if version, ok := versionExpr.(string); ok {
		opts = append(opts, pulumi.Version(version))
	} else {
		e.error(t.CallOpts.Version, "version must be a string value")
	}
}
```

iwahbe: Let's make sure that there is an issue to update the documentation on this issue then.

pkg/pulumiyaml/run.go
```diff
@@ -240,11 +242,22 @@ func (r *Runner) setDefaultProviders() {
 			}
 
 			if v.Options.Provider == nil {
-				if v.Options.Version != nil && v.Options.Version.Value != defaultProviderInfo.version.Value {
-					ctx.addErrDiag(node.Key.Syntax().Syntax().Range(),
-						"Version conflicts with the default provider version",
-						fmt.Sprintf("Try removing this option on resource \"%s\"", k))
+				stringValue := func(expr ast.Expr) string {
```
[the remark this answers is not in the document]
Zaid-Ajaj: Updated the `setDefaultProviders` function to take an evaluator and actually get the value from the visited resources ✅

pkg/pulumiyaml/run_options_test.go
Zaid-Ajaj: Added an assertion that the registered provider has the version evaluated correctly ✅

pkg/pulumiyaml/analyser.go
```diff
@@ -563,11 +565,20 @@ func (tc *typeCache) assertTypeAssignable(ctx *evalContext, from ast.Expr, to sc
...
+	var version *semver.Version
+	if v.Options.Version != nil {
+		switch versionValue := v.Options.Version.(type) {
+		case *ast.StringExpr:
+			parsedVersion, err := ParseVersion(versionValue)
+			if err != nil {
+				ctx.error(v.Type, fmt.Sprintf("unable to parse resource %v provider version: %v", k, err))
+				return true
+			}
+			version = parsedVersion
```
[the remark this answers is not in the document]
Zaid-Ajaj: Updated such that the type checker computes the type of the `Version` and makes sure it is assignable to `schema.StringType`. Then at runtime when we evaluate `Version`, if it has outputs, we return an error.

pkg/pulumiyaml/run.go
Zaid-Ajaj: Yeah added a type-check for `Version` such that it must be assignable to string, so removed this one ✅

Frassle (approved):
I think we need a test of what happens if you try and set the expression to something that resolves to `unknown` at preview time. But otherwise this looks ok.

iwahbe:
@Zaid-Ajaj Do you still want to carry this PR forward?

PR #511, opened 2023-10-17
use property values for config when available

tgummerer:
In https://github.com/pulumi/pulumi/pull/14273 we start passing through config values as PropertyValues.  Use those in pulumi-yaml, so we have proper type information that we can then use to process the config.

Fixes https://github.com/pulumi/pulumi-yaml/issues/453.


pkg/tests/integration_test.go
```diff
@@ -33,8 +33,7 @@ func TestTypeCheckError(t *testing.T) {
 func TestMismatchedConfigType(t *testing.T) {
 	testWrapper(t, integrationDir("mismatched-config-type"), ExpectFailure, StderrValidator{
 		f: func(t *testing.T, stderr string) {
-			assert.Contains(t, stderr,
-				`config key "foo" cannot have conflicting types boolean, integer`)
+			assert.Regexp(t, `config key "foo" cannot have conflicting types boolean, (integer|number)`, stderr)
```
tgummerer, line 36: Since PropertyValue's are always passed as floats we end up having a number type here instead of integer.  I don't think it makes much of a difference in practice, since this is all yaml anyway, but if I'm missing something we can probably try to cast it to integers.

tgummerer:
I think this is ready for a review, though CI unfortunately won't pass before we release a pulumi version that includes https://github.com/pulumi/pulumi/commit/80acb16a54efeea47f1f93b724333a44ff3ebd43, as I don't think we currently have a good way to install a non-released version of pulumi.  If we do I'd be happy to change to using that though.

go.mod
```diff
@@ -8,14 +8,14 @@ require (
...
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/hashicorp/hcl/v2 v2.16.2
+	github.com/hashicorp/hcl/v2 v2.17.0
 	github.com/hexops/autogold v1.3.0
 	github.com/iancoleman/strcase v0.2.0
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/pkg/v3 v3.78.1
-	github.com/pulumi/pulumi/sdk/v3 v3.78.1
+	github.com/pulumi/pulumi/pkg/v3 v3.78.2-0.20231020104416-80acb16a54ef
```
Frassle, line 15: Can update this to the latest pkg&sdk versions.

tgummerer: Done :heavy_check_mark:

pkg/pulumiyaml/run_test.go
```diff
@@ -1868,59 +1864,6 @@ resources:
 	})
 }
 
-func TestGetPulumiConfNodes(t *testing.T) {
```
Frassle, line 1871: Should we not be testing  getConfNodesFromMap as a replacement?

tgummerer: Yeah probably.  Done :heavy_check_mark:

Frassle (approved):

PR #517, opened 2023-10-30
avoid flakyness in config nodes test

tgummerer:
The ordering of the nodes we get back from getConfNodesFromMap is not guaranteed to be stable.  Let's make sure we don't have a flaky test by sorting the result before comparing it to our expected result.

### DIFF
--- a/pkg/pulumiyaml/run_test.go
+++ b/pkg/pulumiyaml/run_test.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


pkg/pulumiyaml/run_test.go
```diff
@@ -1975,7 +1991,7 @@ func TestGetConfNodesFromMap(t *testing.T) {
 		t.Run(tt.project, func(t *testing.T) {
 			t.Parallel()
 			result := getConfNodesFromMap(tt.project, tt.propertymap)
-			assert.Equal(t, tt.expected, result)
+			assert.Equal(t, sortConfigNodes(tt.expected), sortConfigNodes(result))
```
Frassle, line 1994: Might be simpler to use `assert.ElementsMatch` rather than sorting.

tgummerer: TIL, that's much better indeed.

Frassle (approved):

Issue #518, opened 2023-11-09
secret config does not work as of v1.4.0


pgavlin:
I am reasonably certain that the recent changes to consume config as property values were available has broken support for secret config. Consider the stack + program below:

```yaml
# Stack config
config:
  static-secrets:my-secret:
    secure: AAAB... # plaintext value is password
```

```yaml
# Pulumi program
name: static-secrets
runtime: yaml
config:
  my-secret:
    type: string
outputs:
  my-secret: ${my-secret}
```

With `pulumi` v3.92.0 and `pulumi-yaml` v1.4.0, updating this stack leaves the `my-secret` output with the value `{}`:

```console
static-secrets ❯ pulumi stack                                 
Current stack is dev:
    Owner: pat-pulumi-corp
    Last updated: 2 seconds ago (2023-11-09 14:42:43.283559 -0800 PST)
    Pulumi version used: 3.92.0
Current stack resources (1):
    TYPE                 NAME
    pulumi:pulumi:Stack  static-secrets-dev

Current stack outputs (1):
    OUTPUT     VALUE
    my-secret  {}

More information at: https://app.pulumi.com/pat-pulumi-corp/static-secrets/dev

Use `pulumi stack select` to change stack; `pulumi stack ls` lists known ones
```

 With `pulumi-yaml` v1.3.0, the update correctly sets the value of `my-secret` to `password`:

```console
static-secrets ❯ pulumi stack                                               
Current stack is dev:
    Owner: pat-pulumi-corp
    Last updated: 5 seconds ago (2023-11-09 14:41:12.667638 -0800 PST)
    Pulumi version used: 3.92.0
Current stack resources (1):
    TYPE                 NAME
    pulumi:pulumi:Stack  static-secrets-dev

Current stack outputs (1):
    OUTPUT     VALUE
    my-secret  password

More information at: https://app.pulumi.com/pat-pulumi-corp/static-secrets/dev

Use `pulumi stack select` to change stack; `pulumi stack ls` lists known ones
```

PR #519, opened 2023-11-10
fix secrets when config property map is passed

tgummerer:
Currently we're not treating secrets that are passed through a property map correctly.  Fix this by extracting them from the property map, and then wrapping them again in the format the yaml library expects internally.

Fixes https://github.com/pulumi/pulumi-yaml/issues/518

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -935,10 +935,11 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
 			markSecret = true
 		}
 	case configNodeProp:
+		v := intm.value()
 		if intm.v.IsSecret() {
-			return pulumi.ToSecret(intm.v.SecretValue().Element.V), true
+			v = pulumi.ToSecret(intm.v)
 		}
-		return intm.value(), true
+		return v, true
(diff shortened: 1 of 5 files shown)


dixler (approved):
LGTM

pkg/pulumiyaml/run.go
```diff
@@ -935,6 +935,9 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
 			markSecret = true
 		}
 	case configNodeProp:
+		if intm.v.IsSecret() {
```
Frassle, line 938: Ah this is expecting an Output value, not a PropertyValue. Good ol' `interface{}` failing to catch any errors again.
Shouldn't this just match the code below though pretty much of?
```
v := intm.value()
if intm.v.IsSecret() {
	v = pulumi.ToSecret(v)
}
return v, true
```

tgummerer: What you have here doesn't quite work, because `intm.value()` already unwraps the property value (it essentially returns `intm.v.V`), and from my testing we end up with the same result then.  using `pulumi.ToSecret(intm.v)` does work, so we can go with that.

Issue #477, opened 2023-08-01
codegen: logical-name test is broken


abhinav:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The logical-name test is broken for this repository using Pulumi v3.76.

```
--- FAIL: TestGenerateProgram (2.41s)
    --- FAIL: TestGenerateProgram/logical-name (0.00s)
        gen_program_test.go:217:
                Error Trace:    /Users/abg/src/pulumi/yaml/pkg/pulumiyaml/codegen/gen_program_test.go:217
                                                        /Users/abg/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.76.1/codegen/testing/test/program_driver.go:635
                Error:          Received unexpected error:
                                1 error occurred:
                                        * ../testing/test/testdata/logical-name-pp/yaml/logical-name.yaml:2,3-20: missing required configuration variable 'configLexicalName'; run `pulumi config` to set;

                Test:           TestGenerateProgram/logical-name
FAIL
exit status 1
FAIL    github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/codegen    2.453s
```

Discovered in #474

Zaid-Ajaj:
Fixed by https://github.com/pulumi/pulumi-yaml/pull/546

PR #546, opened 2024-01-18
Support (logical) name for resources and config variables

### LINKED ISSUE #539: Broken YAML generated by the converter
### What happened?

There is a surprisingly pesky problem with YAML converter generating invalid YAML with repeated "example:" keys. During programhunt hackathon project with @wongstein and @mnlumi we noted that this is an extremely prevalent error for registry examples (in GCP and AWS), with 255 instances out of 1-2K samples:

```
error_type
resource, variable, or config value .. not found   780
pulumi stack init                                  260
pulumi stack init: "example:"                      255
unknown                                            170
...
```

https://github.com/pulumi/programhunt/blob/main/REPORT.md#registry-example-quality

More recently double-checking pulumi-azuread rollout results, we were able to further pin it down to something specifically introduced by adding the TF converter (pulumi convert) to the examples pipeline. Looks like it's either an intrinsic issue or we're not configuring it right to emit the "good" resource names.

It appears that Java is affected as well https://github.com/pulumi/pulumi-converter-terraform/issues/88 but the issue is not nearly as important there as the new program, while less aesthetically pleasing, still has a chance to work.

### Example

Original HCL:

```
data "azuread_client_config" "current" {}

resource "azuread_user" "example" {
  display_name        = "J Doe"
  owners              = [data.azuread_client_config.current.object_id]
  password            = "notSecure123"
  user_principal_name = "jdoe@hashicorp.com"
}

resource "azuread_group" "example" {
  display_name     = "MyGroup"
  owners           = [data.azuread_client_config.current.object_id]
  security_enabled = true

  members = [
    azuread_user.example.object_id,
    /* more users */
  ]
}
```

Converter generates this. Note how "example" cannot be listed as a dictionary key twice, so this is not something that gets accepted by the raw YAML parser even before Pulumi has a chance to do semantic processing. This is broken:

```yaml
resources:
  example:
    type: azuread:User
    properties:
      displayName: J Doe
      owners:
        - ${current.objectId}
      password: notSecure123
      userPrincipalName: jdoe@hashicorp.com
  example:
    type: azuread:Group
    properties:
      displayName: MyGroup
      owners:
        - ${current.objectId}
      securityEnabled: true
      members:
        - ${example.objectId}
variables:
  current:
    fn::invoke:
      Function: azuread:getClientConfig
      Arguments: {}
```

Prior to introducing the converter the result that used to be generated is this:

```yaml
resources:
  exampleUser:
    type: azuread:User
    properties:
      displayName: J Doe
      owners:
        - ${current.objectId}
      password: notSecure123
      userPrincipalName: jdoe@hashicorp.com
  exampleGroup:
    type: azuread:Group
    properties:
      displayName: MyGroup
      owners:
        - ${current.objectId}
      securityEnabled: true
      members:
        - ${exampleUser.objectId}
variables:
  current:
    fn::invoke:
      Function: azuread:getClientConfig
      Arguments: {}
```

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Zaid-Ajaj:
### Description

Fixes #539, #477 

This PR adds support for resources and config variables to have a _logical name_ other than their lexical name which is used to reference the resources or variables in the YAML program. 

Added as a top-level field:

```yaml
name: simple-yaml
runtime: yaml
resources:
  bucket:
    type: aws:s3:Bucket
    name: my-bucket
    properties:
      website:
        indexDocument: index.html
```

Where the `bucket` is the lexical name that is used for the program evaluator and `my-bucket` is the name that is used when registering the resource. In TypeScript it would look like this:
```typescript
const bucket = new Bucket("my-bucket", ...)
```

Similarly, configuration variables can now have a _logical_ name as well where at runtime, it will be the key used to query the config data provided by Pulumi CLI whereas their lexical name is just how they are defined and referenced in the program. 

Program-gen for YAML has been updated. Take this PCL:
```
config configLexicalName string {
  __logicalName = "cC-Charlie_charlie.😃⁉️"
}

resource resourceLexicalName "random:index/randomPet:RandomPet" {
  // not necessarily a valid logical name, just testing that it passes through to codegen unmodified
  __logicalName = "aA-Alpha_alpha.🤯⁉️"

  prefix = configLexicalName
}

output outputLexicalName {
  __logicalName = "bB-Beta_beta.💜⁉"
  value = resourceLexicalName.id
}
```
Becomes:
```yaml
configuration:
  configLexicalName:
    type: string
    name: "cC-Charlie_charlie.\U0001F603⁉️"
resources:
  resourceLexicalName:
    type: random:RandomPet
    name: "aA-Alpha_alpha.\U0001F92F⁉️"
    properties:
      prefix: ${configLexicalName}
outputs:
  "bB-Beta_beta.\U0001F49C⁉": ${resourceLexicalName.id}
```
> Outputs just keep using their logical name because their lexical name cannot be referenced anywhere

### DIFF
--- a/pkg/pulumiyaml/ast/template.go
+++ b/pkg/pulumiyaml/ast/template.go
@@ -272,6 +272,7 @@ type ConfigParamDecl struct {
 	declNode
 
 	Type    *StringExpr
+	Name    *StringExpr
 	Secret  *BooleanExpr
 	Default Expr
 	Value   Expr
@@ -281,19 +282,20 @@ func (d *ConfigParamDecl) recordSyntax() *syntax.Node {
 	return &d.syntax
 }
 
-func ConfigParamSyntax(node *syntax.ObjectNode, typ *StringExpr,
+func ConfigParamSyntax(node *syntax.ObjectNode, typ *StringExpr, name *StringExpr,
 	secret *BooleanExpr, defaultValue Expr) *ConfigParamDecl {
 
 	return &ConfigParamDecl{
 		declNode: decl(node),
 		Type:     typ,
+		Name:     name,
 		Secret:   secret,
 		Default:  defaultValue,
 	}
 }
 
-func ConfigParam(typ *StringExpr, defaultValue Expr, secret *BooleanExpr) *ConfigParamDecl {
-	return ConfigParamSyntax(nil, typ, secret, defaultValue)
+func ConfigParam(typ *StringExpr, name *StringExpr, defaultValue Expr, secret *BooleanExpr) *ConfigParamDecl {
+	return ConfigParamSyntax(nil, typ, name, secret, defaultValue)
 }
 
 type ResourceOptionsDecl struct {
@@ -411,6 +413,7 @@ type ResourceDecl struct {
 	declNode
 
 	Type            *StringExpr
+	Name            *StringExpr
 	DefaultProvider *BooleanExpr
 	Properties      PropertyMapDecl
 	Options         ResourceOptionsDecl
@@ -423,14 +426,15 @@ func (d *ResourceDecl) recordSyntax() *syntax.Node {
 
 // The names of exported fields.
 func (*ResourceDecl) Fields() []string {
-	return []string{"type", "defaultprovider", "properties", "options", "get"}
+	return []string{"type", "name", "defaultprovider", "properties", "options", "get"}
 }
 
-func ResourceSyntax(node *syntax.ObjectNode, typ *StringExpr, defaultProvider *BooleanExpr,
+func ResourceSyntax(node *syntax.ObjectNode, typ *StringExpr, name *StringExpr, defaultProvider *BooleanExpr,
 	properties PropertyMapDecl, options ResourceOptionsDecl, get GetResourceDecl) *ResourceDecl {
 	return &ResourceDecl{
 		declNode:        decl(node),
 		Type:            typ,
+		Name:            name,
 		DefaultProvider: defaultProvider,
 		Properties:      properties,
 		Options:         options,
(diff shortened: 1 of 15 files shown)


pkg/pulumiyaml/ast/template.go
```diff
@@ -423,23 +426,30 @@ func (d *ResourceDecl) recordSyntax() *syntax.Node {
 
 // The names of exported fields.
 func (*ResourceDecl) Fields() []string {
-	return []string{"type", "defaultprovider", "properties", "options", "get"}
+	return []string{"type", "logicalname", "defaultprovider", "properties", "options", "get"}
```
Frassle, line 429: I wonder if the user facing name for this should just be "name"?

Frassle:
> Outputs just keep using their logical name because their lexical name cannot be referenced anywhere

Why do we even support __logicalName on outputs in PCL? I don't think any language supports referencing of outputs, this looks like something we should clean up.

Frassle:
> Why do we even support __logicalName on outputs in PCL?

My first thought was maybe block labels couldn't be arbitary strings, and that's true for the _first_ label but that's always going to be "output". The second block label is fine to be a string literal: https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#structural-elements

We should sanitize this.

pkg/pulumiyaml/ast/template.go
Zaid-Ajaj: Is this a very common use-case? 🤔 YAML already allows snake-case names and other weird characters in the key of the resource. Though `name` does make sense here as well.

Zaid-Ajaj:
> Why do we even support __logicalName on outputs in PCL? I don't think any language supports referencing of outputs, this looks like something we should clean up.

Indeed 💯

pkg/pulumiyaml/ast/template.go
Frassle: Not really but in all the other sdks we just call this ctor parameter "name". I think it would be very consistent to do that here.

Zaid-Ajaj: Alright, will change it to `name` 👍 same for `config` blocks

Zaid-Ajaj: Done ✅

Frassle (approved):
LGTM great work for getting this done so quickly, will really help unblock providers. Glad it wasn't too gnarly.

AaronFriel:
@Frassle outputs have logical names because they can have names that are not legal identifiers using the `export` facility.

https://github.com/pulumi/pulumi/pull/9464#issuecomment-1109026906

PR #549, opened 2024-02-02
Enable YAML conformance tests

Zaid-Ajaj:
This PR is initial work for enable YAML conformance tests, implements `GeneratePackage`, `GenerateProgram` and `Pack` in the language plugin. Correctly propagates `RootDirectory` as the directory where the YAML templates should be loaded from. 

Implementing local dependencies was a bit tricky, I went the route of adding support for it in program-gen where a package dependency that version `{version}` becomes a default provider resource in the YAML program itself:
```yaml
resources:
  default-simple-provider:
    type: pulumi:providers:simple
    defaultProvider: true
    options:
      version: 1.0.0
  res:
    type: simple:Resource
    properties:
      value: true
```
This however fails today in YAML runtime because it tries to load the `simple` package from an existing binary or tries to download it but that doesn't work because in conformance tests these local packages have an in-memory implementation. 

YAML runtime does its own plugin loading to type-check the template which blocks conformance tests. We need to be able to load schemas from the engine and simply connect to a schema loader service from YAML runtime


Frassle:
> Implementing local dependencies was a bit tricky,

So there's plugin querying which unquestionably needs to return the right thing, and dependency querying which is useful if it returns the right thing but for yaml it doesn't _really_ have dependencies.

So maybe we make conformance testing allow a language plugin to return NotImplemented for GetProgramDependencies.

PR #576, opened 2024-05-10
Add built-in function to convert RFC3339 to Unix timestamps

SayakMukhopadhyay:
Some providers outputs unix timestamps and some other RFC 3339 where some providers input unix and others RFC 3339 and I need to convert between the 2 formats when passing outputs to inputs. I am using YAML only as I want to keep using declarative syntax. Hence, based on the discussion I had in the community slack (ref: https://pulumi-community.slack.com/archives/C037PV12W6L/p1715335945150989) with @t0yv0 .

I don't have much experience with the inner workings of this repo and am mostly using `toBase64` as an example to write this.

### DIFF
--- a/pkg/pulumiyaml/ast/expr.go
+++ b/pkg/pulumiyaml/ast/expr.go
@@ -735,6 +735,19 @@ func ReadFileSyntax(node syntax.Node, name *StringExpr, path Expr) *ReadFileExpr
 	}
 }
 
+type RFC3339ToUnixExpr struct {
+	builtinNode
+
+	Value Expr
+}
+
+func RFC3339ToUnixSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr) *RFC3339ToUnixExpr {
+	return &RFC3339ToUnixExpr{
+		builtinNode: builtin(node, name, args),
+		Value:       args,
+	}
+}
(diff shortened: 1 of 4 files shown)


Frassle:
Rather than adding this as a builtin to yaml it looks like it would fit as a function in pulumi-std (https://github.com/pulumi/pulumi-std). We're trying to prefer helper functions in there rather than specifically in yaml as it allows other languages to benefit from them as well.

t0yv0:
Thanks @Frassle  - I apologize for directing to the wrong repo, I wasn't quite up to date on this.

SayakMukhopadhyay:
@Frassle Thanks for directing me to the correct place. I will close this PR and create a new PR from the std package. Just one question, the std package isn't in the registry right? Although there is some docs in the repo itself, I don't see any docs in the pulumi website or it might be an oversight on my part.

t0yv0:
Good call, we'll have a look! https://github.com/pulumi/pulumi-std/issues/55 thanks!

SayakMukhopadhyay:
In the meantime, I will work on adding the functions to `std` anyway since I can install the provider just fine. Thank you for the quick responses. I will probably be able to create a draft PR before bed.

PR #578, opened 2024-06-03
feat: add PULUMI_STACK for compiler run enviroment

phramz:
This PR adds a new environment variable `PULUMI_STACK` to the compiler run. 

The possibility to determine the stack from the compiler makes it a lot more flexibel when dealing with stack specific resources.

Here an example (I'm using Jinja2 though the issue itself is not compiler specific):
```yaml
# Pulumi.yaml
runtime:
  name: yaml
  options:
    compiler: jinja2 --format=yaml Pulumi.yaml values.yaml
    
 resources:
  # stack specific resources for stage
  # {% if 'stage' == 'PULUMI_STACK'|getenv %}{% include 'templates/mailcatcher.yaml.j2' %}{% endif %}
  
  # common resources
```
A more complex scenario would be to add HA resources to a production stack where an integration system is zonal set up to reduce cloud costs.

My approach seems a bit naive to be honest, but I couldn't figure out a better way without breaking interfaces. Also this solution only works from the `Run`-call context, so `GetRequiredPlugins` might not be accurate when plugins are stack specific.

WDYT?

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,5 @@
 ### Improvements
 
+- Pass `PULUMI_STACK`, `PULUMI_ORGANIZATION`, `PULUMI_PROJECT` and `PULUMI_CONFIG` as environment variable to compiler process. 
+
 ### Bug Fixes
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -46,7 +46,7 @@ func Load() (*ast.TemplateDecl, syntax.Diagnostics, error) {
 	return LoadDir(".")
 }
 
-func LoadFromCompiler(compiler string, workingDirectory string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
+func LoadFromCompiler(compiler string, workingDirectory string, env []string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
 	var diags syntax.Diagnostics
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
@@ -60,6 +60,11 @@ func LoadFromCompiler(compiler string, workingDirectory string) (*ast.TemplateDe
 	cmd.Dir = workingDirectory
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+	cmd.Env = append(env, os.Environ()...)
+	for _, duplicate := range conflictingEnvVars(cmd.Env) {
+		diags = append(diags, syntax.Warning(nil, fmt.Sprintf("environment variable %v is already set: compiler %v will not override conflicting environment variables", duplicate, name), ""))
+	}
+
 	err = cmd.Run()
 	if err != nil {
 		return nil, nil, fmt.Errorf("error running compiler %v: %v, stderr follows: %v", name, err, stderr.String())
@@ -72,6 +77,21 @@ func LoadFromCompiler(compiler string, workingDirectory string) (*ast.TemplateDe
 	diags.Extend(tdiags...)
 
 	return template, tdiags, err
+}
+
+func conflictingEnvVars(env []string) []string {
+	envMap := make(map[string]uint)
+	var duplicates []string
+	for _, e := range env {
+		key := strings.Split(e, "=")[0]
+		if cnt, exists := envMap[key]; exists && cnt <= 1 {
+			duplicates = append(duplicates, key)
+		}
+
+		envMap[key]++
+	}
+
+	return duplicates
 }
 
 // Load a template from the current working directory.
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -18,6 +18,8 @@ package server
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"os"
 
 	pbempty "github.com/golang/protobuf/ptypes/empty"
@@ -53,8 +55,8 @@ func NewLanguageHost(engineAddress, tracing string, compiler string) pulumirpc.L
 	}
 }
 
-func (host *yamlLanguageHost) loadTemplate() (*ast.TemplateDecl, syntax.Diagnostics, error) {
-	if host.template != nil {
+func (host *yamlLanguageHost) loadTemplate(compilerEnv []string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
+	if host.template != nil && host.compiler == "" {
 		return host.template, host.diags, nil
 	}
 
@@ -64,7 +66,7 @@ func (host *yamlLanguageHost) loadTemplate() (*ast.TemplateDecl, syntax.Diagnost
 	if host.compiler == "" {
 		template, diags, err = pulumiyaml.Load()
 	} else {
-		template, diags, err = pulumiyaml.LoadFromCompiler(host.compiler, "")
+		template, diags, err = pulumiyaml.LoadFromCompiler(host.compiler, "", compilerEnv)
 	}
 	if err != nil {
 		return nil, diags, err
@@ -82,7 +84,7 @@ func (host *yamlLanguageHost) loadTemplate() (*ast.TemplateDecl, syntax.Diagnost
 func (host *yamlLanguageHost) GetRequiredPlugins(ctx context.Context,
 	req *pulumirpc.GetRequiredPluginsRequest,
 ) (*pulumirpc.GetRequiredPluginsResponse, error) {
-	template, diags, err := host.loadTemplate()
+	template, diags, err := host.loadTemplate(nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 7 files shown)


pkg/pulumiyaml/run.go
```diff
@@ -60,6 +60,7 @@ func LoadFromCompiler(compiler string, workingDirectory string) (*ast.TemplateDe
 	cmd.Dir = workingDirectory
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+	cmd.Env = env
```
tgummerer, line 63: I don't think this is backwards compatible to do.  Currently the compiler would inherit the environment from the process, but now we're only passing `PULUMI_STACK` in.  I don't think that is quite right, and we probably want to merge the processes environment and the environment we create internally.

tgummerer (commented):
Thanks for your contribution!  This seems reasonable to me at a glance.  Ideally we would also have a test showing that this works.

Also curious if others from @pulumi/platform or @AaronFriel have any opinion on this, or other suggestions of how to implement this.

pkg/pulumiyaml/run.go
Frassle: Yup, this needs to get the existing environment.

pkg/server/server.go
```diff
@@ -121,7 +121,8 @@ func (host *yamlLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest
 		}
 	}
 
-	template, diags, err := host.loadTemplate()
+	compilerEnv := []string{fmt.Sprintf(`PULUMI_STACK=%s`, req.GetStack())}
```
Frassle, line 124: There's other values we know at compile time like the org name, project, config, etc. Should we be setting envvars for all of these? Should we just have one big envvar with a JSON like object for all of them?

pkg/pulumiyaml/run.go
phramz: Good point! Updated the PR and now mergin the `os.Environ()`

pkg/server/server.go
phramz: Added org, project and config (as JSON) to the environment

phramz:
>  Ideally we would also have a test showing that this works.

Regarding a test ... I'm not sure where and how to ideally implement such a test. Would be an acceptance test unless we can mock the `exec` package somehow (probably a little refactoring). 

Any thoughts or ideas on this @tgummerer

tgummerer:
I think an integration test would be fine here, maybe creating two different stacks with different outputs based on the environment variables you are introducing here?  `pkg/tests/integration_test.go` has some examples of current integration tests.

CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,5 @@
 ### Improvements
 
+- Pass `PULUMI_STACK`, `PULUMI_ORGANIZATION`, `PULUMI_ORGANIZATION` and `PULUMI_CONFIG` as environment variable to compiler process.
```
Frassle, line 3:
```suggestion
- Pass `PULUMI_STACK`, `PULUMI_PROJECT`, `PULUMI_ORGANIZATION` and `PULUMI_CONFIG` as environment variable to compiler process. 
```

phramz: done

phramz:
> I think an integration test would be fine here, maybe creating two different stacks with different outputs based on the environment variables you are introducing here? `pkg/tests/integration_test.go` has some examples of current integration tests.

@tgummerer I'm having a hard time setting up the integration tests to run locally. Obviously Pulumi is not using my updated binary as language plugin. So my changes don't have any effect during the test. Is there a doc or could you give me a hint on how to run them locally.

I already pushed an unfinished test PTAL.

Installing the binary will give me an error:
```
error: the yaml language plugin is bundled with Pulumi, and cannot be directly installed with this command. If you need to reinstall this plugin, reinstall Pulumi via your package manager or install script.
```

tgummerer:
> Installing the binary will give me an error:

How exactly were you trying to install the binary?  I believe you should be able to run `make install` to install the binary (and then prepend `$GOBIN` to the `$PATH`.)  There's also a `make test` Makefile target that should run all tests.  Does that not work for you?

pkg/pulumiyaml/run.go
```diff
@@ -60,6 +60,7 @@ func LoadFromCompiler(compiler string, workingDirectory string) (*ast.TemplateDe
 	cmd.Dir = workingDirectory
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+	cmd.Env = append(os.Environ(), env...)
```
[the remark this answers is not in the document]
phramz: According to the docs already present env-vars will be overridden:
https://pkg.go.dev/os/exec#Cmd.Env

So we should be fine

phramz:
> > Installing the binary will give me an error:
> 
> How exactly were you trying to install the binary? I believe you should be able to run `make install` to install the binary (and then prepend `$GOBIN` to the `$PATH`.) There's also a `make test` Makefile target that should run all tests. Does that not work for you?

Thanks! It was indeed a PATH issue having both Pulumi from source and Brew in place.

I finished/fixed the test, fingers crossed it'll pass now.

pkg/pulumiyaml/run.go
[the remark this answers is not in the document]
tgummerer: Ah thanks, I missed that bit in the docs.

The thing I'm still curious about is which should take precedence?  The `PULUMI_*` variables from the users environment or the ones we set internally?  Right now it's the ones set internally, but I wonder if the users environment should take precedence?  

I think I could justify saying "it doesn't matter because you shouldn't be setting environment variables starting with `PULUMI_*` and expect them to do something we never specified anyway", in which case this seems fine.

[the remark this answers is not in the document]
tgummerer: I think the scenario I'm most worried about is if a user is already using a variable with the `PULUMI_` prefix in their program currently, and this would break them.  Thinking some more about it I think I might be more worried about that then the user being confused if they do have something set in their environment and try to use this feature.

Either way I think it would be good to warn the user in this case, you can see an example of how to do that below (https://github.com/pulumi/pulumi-yaml/pull/578/files#diff-228d45f9d585e3feb8e035e1d905d4321203f645dd1a8be21da364c5699ad671R69)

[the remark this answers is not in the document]
phramz: Fair enough 👍

pkg/pulumiyaml/run.go
```diff
@@ -60,6 +60,11 @@ func LoadFromCompiler(compiler string, workingDirectory string) (*ast.TemplateDe
...
```
tgummerer, line 63:
```suggestion
	cmd.Env = append(env, os.Environ()...)
```

I think I would still be more comfortable doing it this way around (and adjusting the warning message).  That would make this change 100% backwards compatible, otherwise I'm a bit worried about that.

pkg/server/server.go
```diff
@@ -18,15 +18,16 @@ package server
...
+	"encoding/json"
+	"fmt"
 	pbempty "github.com/golang/protobuf/ptypes/empty"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/version"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 	"google.golang.org/protobuf/types/known/emptypb"
+	"os"
```
tgummerer, line 30: We'll want to keep the imports from the default library in its own section.

pkg/pulumiyaml/run.go
phramz: Fine by me 👍

tgummerer (approved):
Thank you again, and thanks for the patience while we worked out the best way to also make this backwards compatible.  

I'm gonna leave this PR open until tomorrow just in case anyone else has any comments, but this looks good to me!

phramz:
> Thank you again, and thanks for the patience while we worked out the best way to also make this backwards compatible.
> 
> I'm gonna leave this PR open until tomorrow just in case anyone else has any comments, but this looks good to me!

That went really smooth, kudos and big thanks to you!

artem-zherdiev-ingio:
Hi, could someone provide documentation or an example on how to use it? 

We are currently using our own export PULUMI_STACK_NAME before executing the `compiler: ./compiler` written in golang. So I thought we can now replace custom with `os.LookupEnv(“PULUMI_STACK”)` but it does not work.

I tried this [example](https://github.com/pulumi/pulumi-yaml/blob/316c83a688bf4371180dab9e2765506d1bb6f1b5/pkg/tests/testdata/env-vars/Pulumi.yaml) and it works. But, all of this means that the $PULUMI_STACK variable is only available when pulumi is processing the final yaml, not before, right? I'm a little confused about this snippet of the beginning of this feature, because `{% if 'stage' == 'PULUMI_STACK'|getenv %}{% include 'templates/mailcatcher.yaml.j2' %}{% endif %}` should not work also

Thanks!

phramz:
@artem-zherdiev-ingio 

> But, all of this means that the $PULUMI_STACK variable is only available when pulumi is processing the final yaml, not before, right?

That's correct. Unfortunately on the extension side the stack name is only available for the `Run()` ("final yaml") and not for any other operations like `GetRequiredPlugins()`. This cannot be changed in the yaml-extension since it depends on the API calls from the Pulumi main process. 
That's a big trade-off IMO since compilers have to somehow deal with the fact that the the env-vars could not be around. 

For my use-case the workaround was to have fallback defaults in place when the `PULUMI_STACK` env var is not available so the compiler will still output valid YAML.

> because `{% if 'stage' == 'PULUMI_STACK'|getenv %}{% include 'templates/mailcatcher.yaml.j2' %}{% endif %}` should not work also
  
Besides the given limitations regarding the availability of the env vars the snipped might not work out of the box depending on the Jinja2 installation (the `getenv` filter might not be installed). An alternativ could be using `{% if 'stage' == environ("PULUMI_STACK") %}` in this case.

artem-zherdiev-ingio:
Thank you for your quick reply. Perhaps, it is possible to execute pulumiyaml in our own compiler to get PULUMI_STACK, do custom yaml processing and outputs it to pulumi native compiler. But this looks too complicated.

PR #585, opened 2024-06-05
goreleaser: replace deprecated --skip-validate flag

tgummerer:
This is anothre flag that has been deprecated and removed in v2.0. Replace it with the non-deprecated version.


julienp (approved):
The joy of CI maintenance ❤️

PR #586, opened 2024-06-05
update goreleaser yml files to version 2

tgummerer:
This was the latest issue in
https://github.com/pulumi/pulumi-yaml/actions/runs/9381802513/job/25832183565. Updating the config files to version 2 should hopefully fix this.

I've run `goreleaser check` on them as well, and that passes, so hopefully this the last of it.  The v1.8.0 release went through in the meantime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ obj/
 __pycache__/
 node_modules/
 /coverage
+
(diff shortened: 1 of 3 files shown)


julienp (approved):
🤞

PR #600, opened 2024-07-11
Add tests and fix for StackReference output type checking

AaronFriel:
Ensures that Pulumi YAML programs are able to use StackReference outputs as "any" typed values. Allows "any" values to be assigned to any value, expanding the set of valid programs.

Blocked on:
- https://github.com/pulumi/pulumi/pull/16625

Using a Go workspace with that PR applied to Pulumi, the test suite passes.


Frassle (approved):

pkg/pulumiyaml/analyser.go
```diff
@@ -229,6 +229,10 @@ func (tc *typeCache) isAssignable(fromExpr ast.Expr, to schema.Type) *notAssigna
 		return nil
 	}
 
+	if from == schema.AnyType || to == schema.AnyType {
```
iwahbe, line 232: Allowing an implicit cast from `any` to `T` is unsound. I would have liked to have more discussion about introducing a checked cast instead of just weakening the type checker.

PR #601, opened 2024-07-11
Update changelog for v1.9.1 release

AaronFriel:


lunaris (approved):

CHANGELOG_PENDING.md
```diff
@@ -1,6 +1,3 @@
 ### Improvements
 
 ### Bug Fixes
-
-- Fix Stack Reference output properties to be usable as any type, not only strings inputs.
-  [#600](https://github.com/pulumi/pulumi-yaml/pull/600)
```
justinvp, line 6: Normally, we'd keep this in pending until it's been tagged (and the release created). Then follow-up cleaning it up in a PR after the release is published. But it shouldn't be a problem -- we may just have to edit the release description in GH after it's been created to include this bullet.

Issue #606, opened 2024-07-31
Encountering error/warning when attempting to use strongly typed configuration


zbuchheit:
### What happened?

While attempting to leverage [strongly typed configuration](https://www.pulumi.com/docs/concepts/config/#strongly-typed-configuration) and pulumi-yaml I encountered a warning/error. 

## Repro

1. Run `pulumi preview` with the following setup

### Pulumi.yaml

```yaml
name: pulumi-config-troubleshooting
runtime: yaml
description: A minimal Yaml Pulumi program
config:
  subnets:
    type: array
    items:
      type: string
```

### Pulumi.dev.yaml

```yaml
config:
  pulumi-config-troubleshooting:subnets:
    - subnet2
```

### Warning message

```
Diagnostics:
  pulumi:pulumi:Stack (pulumi-config-troubleshooting-dev):
    Warning: Field 'items' does not exist on Object 'config.subnets'
      on Pulumi.yaml line 7:
       7:     items:
    Existing fields are: 'name', 'type', 'secret', 'value', 'default'
```

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

justinvp:
I'm not personally deeply familiar with the yaml codebase, but from a very quick look, my guess is we'll need to add `Items` to

https://github.com/pulumi/pulumi-yaml/blob/1332be6dae1a81810cd6c1500a3b5eca098ca368/pkg/pulumiyaml/ast/template.go#L271-L279

PR #615, opened 2024-08-21
Parse the items property on config type declarations

julienp:
When parsing a template, parse the `items` property on configuration type declarations to avoid a diagnostic for an unknown field coming from the reflection code. The config section is parsed as part of the template, but not actually used, the config validation is performed by the engine instead.

Fixes https://github.com/pulumi/pulumi-yaml/issues/606

### DIFF
--- a/pkg/pulumiyaml/ast/template.go
+++ b/pkg/pulumiyaml/ast/template.go
@@ -276,6 +276,7 @@ type ConfigParamDecl struct {
 	Secret  *BooleanExpr
 	Default Expr
 	Value   Expr
+	Items   *ConfigParamDecl
 }
 
 func (d *ConfigParamDecl) recordSyntax() *syntax.Node {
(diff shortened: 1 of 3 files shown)


pkg/pulumiyaml/ast/template.go
```diff
@@ -268,6 +268,16 @@ func (d *PropertyMapDecl) parse(name string, node syntax.Node) syntax.Diagnostic
 	return diags
 }
 
+type ConfigItemsDecl struct {
+	declNode
+
+	Type *StringExpr
+}
```
tgummerer, line 275: Do we also need to include values here/is that even supported?

julienp: Ah good spotting! I think this can be arbitrarily nested, we can use `ConfigParamDecl` recursively here, updated the PR!

justinvp (approved):

tgummerer (approved):

PR #616, opened 2024-08-27
Add parameterized provider support to YAML

Frassle:
This adds parameterized provider support to YAML. 

This is done via new "package declaration" files, which are generated by `gen-sdk`. On startup the yaml runtime searches for these files and will use them to register provider packages with the engine. These can include parameterized values, thus supporting parameterized providers.

The schema loader code paths have also been updated to use these package declarations to allow type checking of resources from parameterized packages as well.

Fixes https://github.com/pulumi/pulumi-yaml/issues/626.


go.mod
```diff
@@ -12,13 +14,14 @@ require (
...
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/pkg/v3 v3.130.1-0.20240831132520-d9df47718158
+	github.com/pulumi/pulumi/pkg/v3 v3.130.1-0.20240903214228-52987ac39c3e
 	github.com/pulumi/pulumi/sdk/v3 v3.130.1-0.20240831132520-d9df47718158
 	github.com/spf13/afero v1.9.5
 	github.com/stretchr/testify v1.9.0
 	github.com/zclconf/go-cty v1.13.2
 	google.golang.org/grpc v1.63.2
 	google.golang.org/protobuf v1.33.0
+	gopkg.in/yaml.v2 v2.4.0
```
Frassle, line 24: hmm can probably fix this to just use v3

go.mod
```diff
@@ -12,13 +14,14 @@ require (
 	github.com/hexops/autogold v1.3.0
 	github.com/iancoleman/strcase v0.2.0
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/pkg/v3 v3.130.1-0.20240831132520-d9df47718158
+	github.com/pulumi/pulumi/pkg/v3 v3.130.1-0.20240903214228-52987ac39c3e
 	github.com/pulumi/pulumi/sdk/v3 v3.130.1-0.20240831132520-d9df47718158
```
Frassle, line 18: Should update sdk as well

pkg/pulumiyaml/packages/file.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2024, Pulumi Corporation.  All rights reserved.
```
lunaris, line 1: Note to self -- we should sort the licence headers in this repo

pkg/pulumiyaml/packages/file.go
```diff
@@ -0,0 +1,162 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/blang/semver"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"gopkg.in/yaml.v3"
+)
+
+type ParameterizationDecl struct {
```
lunaris, line 19: Some doc comments on the public interface structs would be nice I reckon.

pkg/pulumiyaml/packages/file.go
```diff
@@ -0,0 +1,162 @@
...
+
+type PackageDecl struct {
+	// PackageDeclarationVersion is the version of the package declaration file.
+	PackageDeclarationVersion int `yaml:"packageDeclarationVersion"`
+	// Name is the name of the plugin.
+	Name string `yaml:"name"`
+	// Version is the version of the plugin.
+	Version string `yaml:"version,omitempty"`
+	// PluginDownloadURL is the URL to download the plugin from.
+	DownloadURL string `yaml:"downloadUrl,omitempty"`
```
lunaris, line 44: Nit/question: do we have any prior art when it comes to acronyms in YAML? E.g. do we favour `downloadUrl` or `downloadURL`? The only one I know is `backend: url:`, but that's the first word so it's lowercase :thinking:

pkg/pulumiyaml/packages/file.go
```diff
@@ -0,0 +1,162 @@
...
+	// Parameterization is the parameterization of the package.
+	Parameterization *ParameterizationDecl `yaml:"parameterization,omitempty"`
+}
+
+func (p *PackageDecl) Valid() bool {
```
lunaris, line 49: Worth changing this to `Validate() error` so we can have better failure reasons?

pkg/pulumiyaml/packages/file.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	return true
+}
+
+// SearchPackageDecls searches the given directory down recursively for package lock .yaml files.
+func SearchPackageDecls(directory string) ([]PackageDecl, error) {
+	if directory == "" {
+		directory = "."
+	}
```
lunaris, line 72: Any benefit to doing this rather than letting the caller just specify `"."`?

pkg/pulumiyaml/packages/file.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		var packageDecl PackageDecl
+		if err := yaml.Unmarshal(data, &packageDecl); err != nil {
+			return fmt.Errorf("unmarshalling %s: %w", path, err)
+		}
+
+		// If the file is not valid skip it
+		if !packageDecl.Valid() {
+			return nil
```
lunaris, line 103: Not sure if we have the infrastructure (and if not, maybe not in this PR) but this feels like a point where we could emit a warning, or something, so that people know if we are skipping over files they think we should be paying attention to. My sense is that you only want errors for files that are lock files but invalid, so perhaps we check to see if the error isn't "no version specified" and only raise then -- anything else is probably just random YAML. Maybe.

pkg/pulumiyaml/packages/file_test.go
```diff
@@ -0,0 +1,39 @@
...
+			Version:                   "1.2",
+			DownloadURL:               "github://api.github.com/pulumiverse",
+		},
+		{
+			PackageDeclarationVersion: 1,
+			Name:                      "base",
+			Parameterization: &ParameterizationDecl{
+				Name:    "pkg",
+				Version: "1.0.0",
+				Value:   "cGtn",
```
lunaris, line 31: This base 64 string is burned into my retina.

pkg/pulumiyaml/run.go
```diff
@@ -343,6 +378,14 @@ func (r *Runner) setDefaultProviders() {
 	contract.IgnoreError(diags)
 }
 
+// Set the runners package descriptors from the templates package decls.
```
lunaris, line 381:
```suggestion
// Set the runner's package descriptors from the templates package decls.
```

pkg/pulumiyaml/run.go
```diff
@@ -343,6 +378,14 @@ func (r *Runner) setDefaultProviders() {
...
+func (r *Runner) setPackageDesciptors() error {
+	// Register package refs for all packages we know upfront
+	packageDescriptors, err := packages.ToPackageDescriptors(r.t.Packages)
+	r.packageDescriptors = packageDescriptors
+	return err
```
lunaris, line 386: Should we always set even if there is an error?

pkg/pulumiyaml/run.go
```diff
@@ -773,7 +825,47 @@ func (e programEvaluator) EvalOutput(r *Runner, node ast.PropertyMapEntry) bool
...
+	packageRefs := make(map[tokens.Package]string)
+	for _, pkg := range r.packageDescriptors {
+		name := pkg.Name
+		// If parametrized use the parametrized name
+		if pkg.Parameterization != nil {
+			name = pkg.Parameterization.Name
+		}
+
+		var parameterization *pulumirpc.Parameterization
+		if pkg.Parameterization != nil {
```
lunaris, line 839: Minor but we could combine these ifs to avoid the double check (cheap as it is). Happy either way. Same is true of `ToPackageDescriptors` earlier too if we feel like it.

pkg/pulumiyaml/run.go
```diff
@@ -1345,7 +1437,9 @@ func (e *programEvaluator) registerResource(kvp resourceNode) (lateboundResource
 
 	// Now register the resulting resource with the engine.
 	if isComponent {
-		err = e.pulumiCtx.RegisterRemoteComponentResource(string(typ), resourceName, untypedArgs(props), res, opts...)
+		typ := tokens.Type(typ)
+		packageRef := e.packageRefs[typ.Package()]
```
lunaris, line 1441: Any cases where this lookup could fail? Same for below.

pkg/server/server.go
```diff
@@ -347,3 +351,87 @@ func (host *yamlLanguageHost) GenerateProgram(
...
+	// The format of the lock file differs based on if this is a parameterized package or not.
+	if pkg.Parameterization == nil {
+		lock.Name = pkg.Name
+		if pkg.Version != nil {
+			lock.Version = pkg.Version.String()
+		}
+		lock.DownloadURL = pkg.PluginDownloadURL
+	} else {
+		lock.Name = pkg.Parameterization.BaseProvider.Name
+		lock.Version = pkg.Parameterization.BaseProvider.Version.String()
```
lunaris, line 400: I know it would be invalid, but are we certain at this point that `Version` is not `nil`? Guessing `BindSpec` guarantees it so no need to be defensive? Or perhaps `Version` is not a pointer here, in which case ignore me :see_no_evil:

lunaris (approved):
Some minor comments but I think this is good to go. One thing I think of is if we'd ever want to put multiple packages in a single lock file, but I think this is fine -- we could introduce a top-level `packages` key and look for that first (optionally bumping the lockfile version).

pkg/pulumiyaml/packages/file.go
Frassle: As with all things at Pulumi it seems inconsistent. In schema files we have "logoUrl" but "pluginDownloadURL". In deployments we have "roleArn" but in state we have "resourceID".

So flip a coin, but "Url" feels more YAML right to me, I don't know why.

pkg/pulumiyaml/packages/file.go
Frassle: Could do now we've got PackageDeclartionVersion. I was worried about err'ing on too many files before.

pkg/pulumiyaml/packages/file.go
Frassle: Not really, lets be strict.

pkg/pulumiyaml/packages/file.go
Frassle: I think we can pass in a diagnostic stream here, I'll have a look.

pkg/pulumiyaml/run.go
Frassle: Probably not

pkg/pulumiyaml/run.go
Frassle: Sure

pkg/pulumiyaml/run.go
Frassle: It can, but the zero value is valid here

pkg/server/server.go
Frassle: It's not a pointer here

pkg/pulumiyaml/run.go
lunaris: Cool :ok_hand:

pkg/server/server.go
lunaris: Oh thank deity -- I was worried I wouldn't get to look like an idiot on this review.

Issue #603, opened 2024-07-22
pulumi ignores `pluginDownloadURL` if program yaml file has resources that reference `config` variables.


MitchellGerdisch:
### What happened?

See comment below - although the code referenced below demonstrates the issue, it is not related to the variables section but simply using config variables in resources.

Using the code here: 
https://gist.github.com/MitchellGerdisch/ac07d6f19cbf9bd2a155b0bfaad0dae5

If the YAML file has a resource with a `pluginDownloadURL` option and a `variables` section, Pulumi will ignore the `pluginDownloadURL` and fail as it looks in the pulumi namespace for the plugin.

### Example

See code here: https://gist.github.com/MitchellGerdisch/ac07d6f19cbf9bd2a155b0bfaad0dae5

### Output of `pulumi about`

CLI          
Version      3.125.0
Go Version   go1.22.5
Go Compiler  gc

Plugins
KIND      NAME  VERSION
language  yaml  unknown

Host     
OS       darwin
Version  13.6.7
Arch     x86_64

This project is written in yaml

Found no resources associated with dev

Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/

No dependencies found

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

MitchellGerdisch:
Actually this is not related to the variables section specifically.
It is because there is a config value that is then referenced in the variables section.
The same thing happens if any resource references a config value.

So this code also demonstrates the problem:

```
name: pequod-yaml-testing
runtime: yaml
description: A minimal Pulumi YAML program
config:
  foo:
    type: string
    default: "bar"
resources: 
  stacksettings:
    type: stackmgmt:StackSettings 
    properties:
      boo: ${foo}
    options:
      pluginDownloadURL: "github://api.github.com/pulumi-pequod/pequod-mlc-stackmgmt"
```

PR #617, opened 2024-08-27
Allow missing nodes in template

julienp:
When a node can not be found, add a missingNode as a placeholder. When walking the graph, VisitMissing can be used to customize the behaviour for these nodes. For typechecking and program evaluation missing nodes are always an error.

During `GetRequiredPlugins` we do not have the program's config available to us. Previously we would abort the graph walk as soon as we hit a missing configuration reference. With this PR, we ignore missing nodes and continue the walk. This allows us detect plugin configuration on resources.

This approach is partially inspired by previous work allowing incomplete templates https://github.com/pulumi/pulumi-yaml/pull/393.

Fixes https://github.com/pulumi/pulumi-yaml/issues/603


tgummerer (approved):

iwahbe (approved):
This looks fine to me.

PR #619, opened 2024-08-29
Test and fix support for local plugins

Frassle:
This updates the pulumi dependency to [d9df4771815894ab6ddaf17b78f6f023c91b7148](https://github.com/pulumi/pulumi/commit/d9df4771815894ab6ddaf17b78f6f023c91b7148) which has a number of fixes for plugin and schema loading.

This allows YAML to correctly load schemas for plugins defined in the Pulumi.yaml file itself. We often use this for integration tests, but we expect users to increasingly make use of this feature in the future as an alternative to dynamic providers.


pkg/tests/testdata/local/Pulumi.yaml
```diff
@@ -0,0 +1,10 @@
+name: local
```
julienp, line 1: Should test provider be listed as a plugin here? Does program test add the plugins to the project for us?

pkg/tests/testprovider/component.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2016-2021, Pulumi Corporation.
```
julienp, line 1: Nit, I think these are copied from pu/pu, since the provider needs to be available at a relative path. Not sure if we need to update the copyright dates here and in the other files.

```suggestion
// Copyright 2024, Pulumi Corporation.
```

julienp (approved):

pkg/tests/testdata/local/Pulumi.yaml
Frassle: Yeh program test adds it for us. We can't add it directly here because program test also copies the test code to a temp dir so a plain relative path doesn't work.

PR #622, opened 2024-09-12
Fix handling invokes that error out when the response contains secrets

Zaid-Ajaj:
### Description

Partially addressing https://github.com/pulumi/pulumi/issues/12710

When the YAML runtime invokes a function that returns a response containing secrets, it would error out because it relies on `Invoke` from the Go SDK which is meant to be used with plain invokes and errors on responses that have secrets. In https://github.com/pulumi/pulumi/pull/17132 we introduced `InvokePackageRaw` which doesn't error and returns a flag to indicate whether the response contained any secrets.

This PR changes the YAML runtime to start using `InvokePackageRaw` and wraps the response with a secret if it contained any. Also adds a test for this use case.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/pulumi/pulumi-yaml
 
-go 1.19
+go 1.21
+
+toolchain go1.22.3
```
Zaid-Ajaj, line 5: `go mod tidy` changed these, not sure if I should discard these changes :thinking:

Zaid-Ajaj: Fails because 
```
github.com/pulumi/esc@v0.9.2-0.20240910221656-328d3204100f requires go@1.21, but 1.19 is requested
```
Do we need to bump Go requirement here to v1.21?

Frassle: Yes

Frassle (approved):

PR #628, opened 2024-09-17
Update workflows to use changie

Frassle:
Update the changelog and release system to use changie.

This matches what we use in the dotnet repo, is a lot nicer to work with than CHANGELOG_PENDING.md files.


tgummerer (approved):
Yes please! :sunglasses:

PR #633, opened 2024-09-17
Trigger another release

Frassle:


.github/workflows/release.yml
```diff
@@ -7,7 +7,7 @@ env:
   GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
 
 jobs:
-  release-pulumi-language-dotnet:
+  release-pulumi-language-yaml:
```
tgummerer, line 10: Hah, actually a useful change here :)

tgummerer (approved):

.github/workflows/release.yml
Frassle: Yeh copy paste fun, noticed it when the job ran

PR #634, opened 2024-09-17
Add environment variable for disabling YAML type checking

VenelinMartinov:
This adds an option to the YAML runtime to disable the type checking. This is controlled by the `PULUMI_YAML_DISABLE_TYPE_CHECKING` environment variable.

This is useful for tests where we need to test type mismatches which could happen in other languages, like Python or Typescript.

Can use some guidance on how to add tests here.

Required for https://github.com/pulumi/pulumi-terraform-bridge/issues/2418

### DIFF
--- a/pkg/pulumiyaml/analyser.go
+++ b/pkg/pulumiyaml/analyser.go
@@ -4,6 +4,7 @@ package pulumiyaml
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
@@ -11,6 +12,7 @@ import (
 	"github.com/hashicorp/hcl/v2"
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 
 	"github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/ast"
 	ctypes "github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/config"
@@ -532,10 +534,15 @@ func hasValidEnumValue(from ast.Expr, to []*schema.Enum) *notAssignable {
 	}
 }
 
+var disableTypeChecking = cmdutil.IsTruthy(os.Getenv("PULUMI_YAML_DISABLE_TYPE_CHECKING"))
+
 // Provides an appropriate diagnostic message if it is illegal to assign `from`
 // to `to`.
 func (tc *typeCache) assertTypeAssignable(ctx *evalContext, from ast.Expr, to schema.Type) {
-	return
+	if disableTypeChecking {
+		return
+	}
+
 	if to == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


pkg/pulumiyaml/analyser.go
```diff
@@ -532,9 +534,15 @@ func hasValidEnumValue(from ast.Expr, to []*schema.Enum) *notAssignable {
...
 }
 
+var disableTypeChecking = cmdutil.IsTruthy(os.Getenv("PULUMI_YAML_DISABLE_TYPE_CHECKING"))
+
 // Provides an appropriate diagnostic message if it is illegal to assign `from`
 // to `to`.
 func (tc *typeCache) assertTypeAssignable(ctx *evalContext, from ast.Expr, to schema.Type) {
+	if disableTypeChecking {
+		return
+	}
```
iwahbe, line 544: I'd probably print a big warning here that this is not something that users should set, the provider may panic, etc.

pkg/pulumiyaml/analyser.go
```diff
@@ -532,9 +534,15 @@ func hasValidEnumValue(from ast.Expr, to []*schema.Enum) *notAssignable {
 	}
 }
 
+var disableTypeChecking = cmdutil.IsTruthy(os.Getenv("PULUMI_YAML_DISABLE_TYPE_CHECKING"))
```
iwahbe, line 537:
```suggestion
var disableTypeChecking = cmdutil.IsTruthy(os.Getenv("PULUMI_DEBUG_YAML_DISABLE_TYPE_CHECKING"))
```

iwahbe (commented):
I'm not sure how much yaml actually works with the wrong types. I think it will still try to do type conversions on occasion.

pkg/pulumiyaml/analyser.go
VenelinMartinov: What's the accepted way to print a warning in the language plugin?

I tried:
```golang
ctx.addWarnDiag(
	from.Syntax().Syntax().Range(),
	"Running with type checking disabled. This is not recommended for production use.",
	"Type checking is disabled. This is a test feature and should not be used in production.",
)
```

but that didn't work

iwahbe: That's what I would recommend. What does "didn't work" mean? Did it just not do anything?

iwahbe (approved):

pkg/pulumiyaml/analyser.go
VenelinMartinov: Yeah, it didn't print anything

t0yv0 (approved):
Beautiful, thanks!

pkg/pulumiyaml/analyser.go
iwahbe: Hmm. That's concerning. I wouldn't block on this.

tgummerer: Adding the warning is probably still a good idea, and opening an issue.  We have other warnings below that we should probably also fix if this doesn't work.

VenelinMartinov: Opened https://github.com/pulumi/pulumi-yaml/issues/638

Frassle (approved):

Issue #604, opened 2024-07-25
Default Provider fails to apply if it's defined below its resource


mikhailshilkov:
### What happened?

I assume the order of YAML resources should not matter. However, this program works:

```yaml
name: aws-yaml
runtime: yaml
resources:
  provider:
    defaultProvider: true
    type: pulumi:providers:aws
    properties:
      region: us-west-2
  alb:
    type: aws:alb:LoadBalancer
    properties:
      tags: 
        Name: test-lb
      name: testing
      subnets: 
        - subnet-eacf3697
        - subnet-939b18f8
```

while the following one fails

```yaml
name: aws-yaml
runtime: yaml
resources:
  alb:
    type: aws:alb:LoadBalancer
    properties:
      tags: 
        Name: test-lb
      name: testing
      subnets: 
        - subnet-eacf3697
        - subnet-939b18f8
  provider:
    defaultProvider: true
    type: pulumi:providers:aws
    properties:
      region: us-west-2
```

with

```
    error: resource or variable named "provider" could not be found
    error: Error registering resource [alb]: no diagnostics
    error: Error registering resource [provider]: no diagnostics
```

### Example

see above

### Output of `pulumi about`

```
CLI          
Version      3.126.0
Go Version   go1.22.5
Go Compiler  gc

Plugins
KIND      NAME  VERSION
resource  aws   unknown
language  yaml  unknown

Host     
OS       darwin
Version  14.5
Arch     arm64

This project is written in yaml
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

brandonpollack23:
Ok after some digging it seems the topological sort is not aware of the dependency of alb on the aws provider here.  Looking into how to pivot on that information if its there and if it isn't figuring out how to add it (perhaps this is an aws provider specific issue but I doubt it)

brandonpollack23:
More progress: if you explicitly set the provider in the config section it works out. the topological sort doesn't work because it doesn't detect "default provider" I bet I could add a special case for that in topological sort to always put that first so any other resources that expect it will work

justinvp:
Fixed by https://github.com/pulumi/pulumi-yaml/pull/652

PR #643, opened 2024-09-24
Fix topological sort to take into account defaultProvider.

brandonpollack23:
[//]: # (BEGIN GIT-PR FOOTER)

The topological sort will not sort default providers first as is, so
this adds in a check to detect it and not allow implicit providers to be
complete until the default provider is declared.

---

* ◻️ #647
* 🐰 #643 (👉[a726ca1d](https://github.com/pulumi/pulumi-yaml/commit/a726ca1d))

### DIFF
--- a/pkg/pulumiyaml/sort.go
+++ b/pkg/pulumiyaml/sort.go
@@ -138,9 +138,16 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 			sorted = append(sorted, node)
 		}
 	}
+
+	var defaultProvider *ast.StringExpr
 	for _, kvp := range t.Resources.Entries {
 		rname, r := kvp.Key.Value, kvp.Value
 		node := resourceNode(kvp)
+
+		// Check if the resource is a default provider
+		if resourceIsDefaultProvider(node) {
+			defaultProvider = node.key()
+		}
 
 		cdiags := checkUniqueNode(intermediates, node)
 		diags = append(diags, cdiags...)
@@ -198,6 +205,7 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 		}
 		if !visited[name.Value] {
 			visiting[name.Value] = true
+
 			for _, mname := range dependencies[name.Value] {
 				if mname.Value == PulumiVarName {
 					continue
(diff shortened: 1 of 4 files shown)


brandonpollack23:
Bug: #604

pkg/tests/integration_test.go
```diff
@@ -133,3 +133,13 @@ func TestLocalPlugin(t *testing.T) {
 		},
 	})
 }
+
+//nolint:paralleltest // uses parallel programtest
+func TestResourceOrderingWithDefaultProvider(t *testing.T) {
```
Frassle, line 138: Does this actually fail without the fix? Doesn't it just use the engine default for the alb and so still preview ok?

pkg/pulumiyaml/sort.go
```diff
@@ -138,10 +138,17 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 			sorted = append(sorted, node)
 		}
 	}
+
+	var defaultProvider *ast.StringExpr
```
Frassle, line 142: There could be multiple default providers for different packages

pkg/tests/testdata/resource-ordering/Pulumi.yaml
```diff
@@ -0,0 +1,17 @@
+name: aws-yaml
+runtime: yaml
+resources:
+  alb:
```
Frassle, line 4: I'd like to see this use two default providers, from two different packages _and_ show that we can have one default provider depend on the output of a resource from the other package. Something like the following:
resA: type: pkg1:idx:typ
resB: type: pkg2:idx:typ
provA: type: pulumi:providers:pkg1
provB: type: pulumi:providers:pkg2
   properties: input: resA.output
   
Should sort as [provA, resA, provB, resB]

pkg/tests/integration_test.go
brandonpollack23: nope!

brandonpollack23: Nope!

pkg/tests/testdata/resource-ordering/Pulumi.yaml
brandonpollack23: WOO did it, here is the output, ill try to add a test for topological sort as well:

  Type                              Name                Plan       Info
 +   pulumi:pulumi:Stack               aws-yaml-teststack  create     2 warnings; 1 message
 +   ├─ pulumi:providers:aws           provider            create
 +   ├─ aws:alb:LoadBalancer           alb                 create
 +   ├─ pulumi:providers:testprovider  testProvider        create
 +   └─ testprovider:index:Echo        echo                create

brandonpollack23: I made it a child PR just in case we wanna roll back one at a time and not squash them, lmk if you prefer to just combine them

PR #651, opened 2024-09-25
update go version in github workflows

brandonpollack23:


tgummerer (approved):
There's also a reference to Go 1.21.x in stage-publish.yml.  Did we not upgrade that intentionally?  Probably deserves a callout in the PR description if so.

> Remote-Ref: brandonpollack23/30597e61

Out of curiosity, what's this?

pkg/pulumiyaml/run.go
```diff
@@ -1064,7 +1064,7 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
 	if errors.Is(err, config.ErrMissingVar) && defaultValue != nil {
 		v = defaultValue
 	} else if err != nil {
-		return e.errorf(intmKey, err.Error())
+		return e.errorf(intmKey, "%s", err.Error())
```
Frassle, line 1067: This isn't the same? If Error has %'s in it doesn't that now trip the formatter that values weren't passed in resulting in MISSING outputs.

pkg/pulumiyaml/codegen/gen_program_test.go
```diff
@@ -132,9 +132,9 @@ func (m FakePackage) IsComponent(typeName pulumiyaml.ResourceTypeToken) (bool, e
 		// note this returns if err *equals* nil.
 		return false, nil
 	}
-	msg := fmt.Sprintf("Unexpected type token in IsComponent: %q", typeName)
-	m.t.Logf(msg)
-	return false, fmt.Errorf(msg)
+
+	m.t.Logf("Unexpected type token in IsComponent: %q", typeName)
```
Frassle, line 136: Is this really better? We've duplicated the text line

Frassle (changes requested):

brandonpollack23:
> There's also a reference to Go 1.21.x in stage-publish.yml. Did we not upgrade that intentionally? Probably deserves a callout in the PR description if so.
> 
> > Remote-Ref: brandonpollack23/30597e61
> 
> Out of curiosity, what's this?

a local tool I use to push stacks of changes as separate PRs adds that

pkg/pulumiyaml/run.go
brandonpollack23: indeed but the old way is a non constant string and golint complains

pkg/pulumiyaml/codegen/gen_program_test.go
brandonpollack23: linters say so, what do you think, should we disable this lint?

brandonpollack23:
> There's also a reference to Go 1.21.x in stage-publish.yml. Did we not upgrade that intentionally? Probably deserves a callout in the PR description if so.
> 
> > Remote-Ref: brandonpollack23/30597e61
> 
> Out of curiosity, what's this?

I don't see stage publish in the repo anywhere....am I blind?

tgummerer:
> I don't see stage publish in the repo anywhere....am I blind?

No I am just outdated.  I had an older checkout lying around, but that script is gone since then.  Never mind me.

.changes/unreleased/Improvements-647.yaml
```diff
@@ -0,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: update go version for github workflows
```
tgummerer, line 3:
```suggestion
body: Update go version for GitHub workflows
```

pkg/pulumiyaml/codegen/gen_program_test.go
tgummerer: I think disabling this linter is reasonable.

pkg/pulumiyaml/run.go
tgummerer: Isn't it the other way around, previously if `err.Error()` has a "%s", it would result in `%!s(MISSING)`, but now it's just a literal `%s`.

E.g. 

```
	fmt.Printf("%s", "test has a %s inside\n")
	fmt.Printf("test has a %s inside")
```

prints

```
test has a %s inside
test has a %!s(MISSING) inside
```

So I think the first one/the way Brandon changed it looks to me more like what we want.

Frassle: Oh it was too early in the morning, I read the diff backwards somehow. Yes this is now correct, the old one was wrong for the reasons given.

pkg/pulumiyaml/codegen/gen_program_test.go
```diff
@@ -64,9 +64,9 @@ func (m FakePackage) ResolveResource(typeName string) (pulumiyaml.ResourceTypeTo
 		OtherThing, OtherModule:
 		return pulumiyaml.ResourceTypeToken(typeName), nil
 	default:
-		msg := fmt.Sprintf("Unexpected type token in ResolveResource: %q", typeName)
-		m.t.Logf(msg)
-		return "", fmt.Errorf(msg)
+
+		m.t.Logf("Unexpected type token in ResolveResource: %q", typeName)
+		return "", fmt.Errorf("Unexpected type token in ResolveResource: %q", typeName)
```
iwahbe, line 69: Let's not introduce duplication if we can help it. 
```suggestion
		msg := fmt.Sprintf("Unexpected type token in ResolveResource: %q", typeName)
		m.t.Logf("%s", msg)
		return "", errors.New(msg)
```

We should do this to all sites where you made this kind of change.

pkg/pulumiyaml/codegen/gen_program_test.go
iwahbe: I don't think we should disable this linter. The lint is useful, and we were doing something bad here. I would suggest a different work-around:

```go
		msg := fmt.Sprintf("Unexpected type token in IsComponent: %q", typeName)
		m.t.Logf("%s", msg)
		return "", errors.New(msg)
```

iwahbe (dismissed):

PR #652, opened 2024-09-25
Fix topological sort to take into account defaultProvider.

brandonpollack23:
The topological sort will not sort default providers first as is, so
this adds in a check to detect it and not allow implicit providers to be
complete until the default provider is declared.

Handle multiple default providers for different packages.

Also added a test to test topological sort works correctly.

### DIFF
--- a/pkg/pulumiyaml/sort.go
+++ b/pkg/pulumiyaml/sort.go
@@ -139,14 +139,16 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 		}
 	}
 
-	var defaultProvider *ast.StringExpr
+	// Map of package name to default provider resource and it's key.
+	defaultProviders := map[string]*ast.StringExpr{}
 	for _, kvp := range t.Resources.Entries {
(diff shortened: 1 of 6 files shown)


Frassle:
> sorry local tool messups, had to recreate PR

Copy the original description to the PR description. Put comments like that in comments, it doesn't need to merge to git history when we merge this.

pkg/pulumiyaml/sort.go
```diff
@@ -206,6 +216,25 @@ func topologicallySortedResources(t *ast.TemplateDecl, externalConfig []configNo
 					return false
 				}
 			}
+
+			if resNode, ok := e.(resourceNode); ok {
+				pkg := ""
+				if resNode.Value.Type != nil {
+					pkg = strings.Split(resNode.Value.Type.Value, ":")[0]
```
Frassle, line 223: Can use `Cut` for cases like this.

justinvp:
Does this fix #604?

FYI, for future reference, if you include `Fixes #604` in the PR description, when the PR is merged, it will automatically close the linked issue.

PR #654, opened 2024-09-29
Rpc wrapper [draft]

im7mortal:
Related to

https://github.com/pulumi/pulumi/pull/17346
https://github.com/pulumi/pulumi/issues/17242

First we need to merge https://github.com/pulumi/pulumi/pull/17346


iwahbe:
Given that this has sat for 2 years, I don't think we need to keep the PR open.

PR #657, opened 2024-10-08
build statically linked binaries with CGO disabled

tgummerer:
In https://github.com/pulumi/pulumi-yaml/pull/632 we switched the build from running on MacOS to Ubuntu.  On MacOS CGO is disabled by default, and switching to Ubuntu means this is turned on.  We however want to build statically linked binaries since they are easier to deal with.  Set that up in the goreleaser.yml explicitly.


lunaris (approved):

justinvp:
Next step is to pull this release into pulumi/pulumi and release a new version of the CLI that includes this.

PR #660, opened 2024-10-08
Use the workflow's GITHUB_TOKEN to create the release

justinvp:
The release failed with the following error:

> 403 The 'Pulumi' enterprise forbids access via a fine-grained personal access tokens if the token's lifetime is greater than 365 days.

Rather than rotating `PULUMI_BOT_TOKEN`, I think we can avoid this going forward by using the workflow's `secrets.GITHUB_TOKEN` with `contents:write` permissions.

Additionally, this change allows dispatching the workflow manually, to allow the workflow to be re-run easily after this is merged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,9 +2,10 @@ on:
   push:
     branches: [ main ]
     paths: [ CHANGELOG.md ]
+  workflow_dispatch: {}
 
-env:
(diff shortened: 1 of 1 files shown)


tgummerer (approved):
I think this is the right thing to do anyway, even if `PULUMI_BOT_TOKEN` is fixed.

PR #669, opened 2024-11-12
Remove unused Load method

Frassle:
Other uses of the load method pass explicit directories, remove this Load helper and just use LoadDir directly to match.


lunaris (approved):

Frassle:
hmm my git must have been a mess... lint says this is used. But it shouldn't be so lets fix that!

tgummerer (approved):

PR #673, opened 2024-11-13
Enable l1-builtin-info test

Frassle:
This enables the l1-builtin-info test. As we're now running multiple templates with the same `yamlLanguageHost` we can't globally cache the template in loadTemplate anymore.

Instead we cache by the program directory. For normal runs there will only be the one program directory, and for conformance test we know each program will be in its own unique directory.


lunaris (approved):
I'm minded for us to spend a bit of time trying to work out why directory caching doesn't work here, since it seems like the fundamental unit that we should be able to capture. Failing that, can we push down to the load and do it at the file level?

Frassle:
> I'm minded for us to spend a bit of time trying to work out why directory caching doesn't work here

Yeh I'll have a fresh look today. Maybe sleeping on it will have helped.

Frassle:
Found the problem with caching, it was only if "compiler" option was set. Fixed to use directory caches and added a comment as to why we don't cache if compiler is set.

pkg/server/server.go
```diff
@@ -84,10 +93,15 @@ func (host *yamlLanguageHost) loadTemplate(directory string, compilerEnv []strin
 	if diags.HasErrors() {
 		return nil, diags, nil
 	}
-	host.template = template
-	host.diags = diags
 
-	return host.template, diags, nil
+	if host.compiler != "" {
```
lunaris, line 97: Might be worth either adding a `shouldCache` method or even just a `cache` method that internalise this check/provide a single place we can document it.

PR #675, opened 2024-11-14
Enable l1-output-string

Frassle:
This required two fixes.

Firstly when reading in a literal string from PCL we need to re-escape any `$${` escape sequences.

Secondly it's valid to have an empty string as a stack export. We fix this by making sure empty strings from PCL write out as `""` rather than nothing, as the latter caused the yaml runtime to think it was an empty/null/missing key rather than an empty string.

### DIFF
--- a/cmd/pulumi-language-yaml/testdata/projects/l1-output-string/Main.yaml
+++ b/cmd/pulumi-language-yaml/testdata/projects/l1-output-string/Main.yaml
@@ -1,9 +1,9 @@
 outputs:
-  empty:
+  empty: ""
   small: Hello world!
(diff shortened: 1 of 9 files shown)


cmd/pulumi-language-yaml/testdata/projects/l1-output-string/Main.yaml
```diff
@@ -1,9 +1,9 @@
 outputs:
```
Frassle, line 1: Accidentally commit the broken version of this file in https://github.com/pulumi/pulumi-yaml/pull/673

PR #679, opened 2024-11-15
Turn on more tests, record remaining failures

Frassle:
This turns on the remaining already passing conformance tests for YAML, and records a reason why the remaining failing tests are failing.


cmd/pulumi-language-yaml/language_test.go
```diff
@@ -172,30 +172,15 @@ func runTestingHost(t *testing.T) (string, testingrpc.LanguageTestClient) {
...
-	"l2-target-up-with-new-dependency":      "TODO",
-	"l2-invoke-dependencies":                "TODO",
-	"l2-large-string":                       "TODO",
-	"l2-provider-grpc-config":               "TODO",
-	"l2-resource-asset-archive":             "TODO",
+	"l1-stack-reference":                    "YAML does not support fn::getOutput",
+	"l2-invoke-options":                     "cannot assign expression",
+	"l2-map-keys":                           "missing expected dependency primitive",
+	"l2-provider-grpc-config-schema-secret": "Detected a secret leak in state",
+	"l2-invoke-secrets":                     "YAML does not support fn::secret",
```
tgummerer, line 179: What does this mean?  YAML has a `fn::secret` here: https://github.com/pulumi/pulumi-yaml/blob/430fa6008462d5bb6f39094a7059c9721cc149e1/pkg/pulumiyaml/ast/expr.go#L782.  Is this different from what we need?  Can we describe that a bit more to avoid confusing others as well?

Frassle: That's just the error message I got when running this test, I didn't dig into it deeper.

PR #684, opened 2024-11-15
Enable l2-resource-asset-archive test

Frassle:
This is a bit more of an involved change to get this test working. Making two major changes.

Firstly, historically YAML would always send absolute paths instead of relative paths to the engine. This was wrong, the engine expects that paths should be relative so that we don't save absolute paths to the state file.

Secondly, YAML would error if you tried to access what it considered "unsafe" paths. That was either paths that clearly accessed outside the Pulumi.yaml directory, or non-constant paths to the remote asset and archive types. This seems overly restrictive, none of our other languages try to protect the user like this and it seems reasonable to expect to be able to do similar filesystem accesses from YAML.


.changes/unreleased/Improvements-684.yaml
```diff
@@ -0,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: Asset and archive paths are passed as is to the engine
```
tgummerer, line 3:
```suggestion
body: Pass asset and archive paths as is to the engine
```

pkg/pulumiyaml/run_test.go
```diff
@@ -1640,39 +1638,6 @@ variables:
...
-// a secret or token.
-func TestReadFileForbidsPathTraversal(t *testing.T) {
-	t.Parallel()
-
-	text := `
-name: test-readfile
-runtime: yaml
-outputs:
-  readme:
-    fn::readFile: ${pulumi.cwd}/../../go.mod # imagine this is /etc/shadow, /var/run/secrets/tokens, etc.
```
tgummerer, line 1659: This made me pause for a second, as it's security related.  I don't think it's in our security model to guard against users doing anything in their programs in general though (that would be really hard to do in other supported languages anyway).  The user running the program needs to be the one to inspect and trust it, so I think removing this restriction is fine.

Frassle: That was my thinking

PR #687, opened 2024-11-18
Enable l2-invoke-options

Frassle:
Pretty easy fix to enable this test, we can just copy the options through directly from PCL (they happen to match in both languages).

This did require updating the pulumi/pkg dependency so that `BindProgram` could correctly bind to a PCL program using invoke options.


Frassle:
Covered by https://github.com/pulumi/pulumi-yaml/pull/678/files

PR #693, opened 2024-11-20
Allow specifying dependencies for output invokes

julienp:
Provider functions that take inputs as arguments, and return an output (aka output invokes), now allow specifying a `dependsOn` option. This allows programs to ensure that invokes are executed after things they depend on, similar to the [`depdendsOn` resource option](https://www.pulumi.com/docs/iac/concepts/options/dependson/).

https://github.com/pulumi/pulumi/discussions/17710

Fixes https://github.com/pulumi/pulumi/issues/17755

### DIFF
--- a/.changes/unreleased/Improvements-693.yaml
+++ b/.changes/unreleased/Improvements-693.yaml
@@ -1,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: Allow specifiying dependencies for output invokes
+time: 2024-11-20T14:58:17.346265+01:00
+custom:
+    PR: "693"
--- a/.changes/unreleased/Improvements-693.yaml
+++ b/.changes/unreleased/Improvements-693.yaml
@@ -1,6 +1,6 @@
 component: runtime
 kind: Improvements
-body: Allow specifiying dependencies for output invokes
+body: Allow specifying dependencies for output invokes
 time: 2024-11-20T14:58:17.346265+01:00
 custom:
     PR: "693"
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -2019,6 +2019,21 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
 			e.error(t.Return, fmt.Sprintf("Unable to evaluate options Provider field: %+v", t.CallOpts.Provider))
 		}
 	}
+	if t.CallOpts.DependsOn != nil {
+		dependsOnOpt, ok := e.evaluateResourceListValuedOption(t.CallOpts.DependsOn, "dependsOn")
+		if ok {
+			var dependsOn []pulumi.Resource
+			for _, r := range dependsOnOpt {
+				if p, ok := r.(poisonMarker); ok {
+					return p, true
+				}
+				dependsOn = append(dependsOn, r.CustomResource())
+			}
+			opts = append(opts, pulumi.DependsOn(dependsOn))
+		} else {
+			e.error(t.Return, fmt.Sprintf("Unable to evaluate options DependsOn field: %+v", t.CallOpts.DependsOn))
+		}
+	}
 	performInvoke := e.lift(func(args ...interface{}) (interface{}, bool) {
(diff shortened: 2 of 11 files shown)


.changes/unreleased/Improvements-693.yaml
```diff
@@ -0,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: Allow specifiying dependencies for output invokes
```
julienp, line 3:
```suggestion
body: Allow specifying dependencies for output invokes
```

pkg/pulumiyaml/run.go
```diff
@@ -2022,27 +2037,30 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
...
 
 		pkgRef := ""
-		secret, err := e.pulumiCtx.InvokePackageRaw(string(functionName), args[0], &result, pkgRef, opts...)
+		secret, deps, err := e.pulumiCtx.InvokePackageRawWithDeps(string(functionName), args[0], &result, pkgRef, opts...)
 		if err != nil {
 			return e.error(t, err.Error())
 		}
 
 		if t.Return.GetValue() == "" {
+			output := pulumi.OutputWithDependencies(context.TODO(), pulumi.Any(result), deps...)
```
lunaris, line 2046: `e.pulumiCtx` has a `.Context()` which might be a good thing to hang this off, rather than `TODO`?

pkg/pulumiyaml/run.go
```diff
@@ -2022,27 +2037,30 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
...
+			return output, true
 		}
 
 		retv, ok := result[t.Return.Value]
 		if !ok {
 			e.error(t.Return, fmt.Sprintf("Unable to evaluate result[%v], result is: %+v", t.Return.Value, t.Return))
 			return e.error(t.Return, fmt.Sprintf("fn::invoke of %s did not contain a property '%s' in the returned value", t.Token.Value, t.Return.Value))
 		}
+
+		output := pulumi.OutputWithDependencies(context.TODO(), pulumi.Any(retv), deps...)
```
lunaris, line 2059: Same thoughts about `context.TODO` here.

pkg/pulumiyaml/run_invoke_test.go
```diff
@@ -216,7 +216,7 @@ runtime: yaml
 	requireNoErrors(t, tmpl, diags)
 }
 
-func TestInvokeReturningSecrets(t *testing.T) {
+func TestInvokeReturningSecretsWithReturn(t *testing.T) {
```
lunaris, line 219: Is this the best name change? I'm guessing one is testing plain, the other outputs? Or is it something different? It's not super clear to me.

lunaris (approved):

pkg/pulumiyaml/run.go
julienp: Yep, there's another call to context.TODO just above in the call to `ResolveFunction`, I'll update that one too.

pkg/pulumiyaml/run_invoke_test.go
julienp: Yeah it's not super clear. This test has a `return` property on the invoke on like 229, so the test covers the `if t.Return.GetValue() == "" {` branch above, whereas the other test returns the whole response, covering the rest of the changes in that file.

go.mod
```diff
@@ -12,8 +12,8 @@ require (
 	github.com/hexops/autogold v1.3.0
 	github.com/iancoleman/strcase v0.2.0
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/pkg/v3 v3.139.0
-	github.com/pulumi/pulumi/sdk/v3 v3.139.0
+	github.com/pulumi/pulumi/pkg/v3 v3.142.1-0.20241210130857-42257d81376e
+	github.com/pulumi/pulumi/sdk/v3 v3.142.1-0.20241210130857-42257d81376e
```
julienp, line 16: TODO @julienp  update on next pu/pu release

PR #698, opened 2024-12-11
Update to pulumi v3.142.0

Frassle:
This added a few tests. Notably l2-resource-secret which I've fixed to pass, and l2-invoke-options-depends-on which I've skipped because I couldn't quickly work out how to fix the output dependency to propagate.


brandonpollack23 (approved):

cmd/pulumi-language-yaml/language_test.go
```diff
@@ -173,6 +173,7 @@ func runTestingHost(t *testing.T) (string, testingrpc.LanguageTestClient) {
 // Add test names here that are expected to fail and the reason why they are failing
 var expectedFailures = map[string]string{
 	"l2-provider-grpc-config-schema-secret": "Detected a secret leak in state",
+	"l2-invoke-options-depends-on":          "Invoke dependency not propagated",
```
julienp, line 176: I have this test implemented in https://github.com/pulumi/pulumi-yaml/pull/693

pkg/pulumiyaml/run_test.go
```diff
@@ -80,6 +81,10 @@ func (m MockPackage) ResolveFunction(typeName string) (FunctionTypeToken, error)
 	return FunctionTypeToken(typeName), nil
 }
 
+func (m MockPackage) IsResourcePropertySecret(typeName ResourceTypeToken, propertyName string) (bool, error) {
+	return m.isResourcePropertySecret(typeName.String(), propertyName)
```
julienp, line 85: There's a test currently failing here, we need to either add an implementation of  `isResourcePropertySecret` to https://github.com/pulumi/pulumi-yaml/blob/0228354d1e833d8cd666d7957e60413a6563061f/pkg/pulumiyaml/run_test.go#L168 or  do something like we do for ResolveFunction/ResolveResource where we return `false, nil` by default here.

pulumi-yaml.sln
```diff
@@ -0,0 +1,644 @@
+﻿
```
tgummerer, line 1: Was this meant file to be committed?

Frassle: No :( vscode

pulumi-yaml.sln
```diff
@@ -0,0 +1,644 @@
...
+Microsoft Visual Studio Solution File, Format Version 12.00
```
tgummerer, line 2: It's back :sweat_smile:  Might be worth adding this to `.gitignore`

Frassle: Yeh lets do that

Frassle:
oddly this seems to be passing on my machine :|

PR #704, opened 2024-12-19
Update release doc

julienp:
This changed now that we don't have a direct dependency on pulumi-yaml in pu/pu anymore.


tgummerer (approved):
:heart:

PR #717, opened 2025-02-18
Make pulumi submodule point to master and override go.mod to point to it.

brandonpollack23:
This makes running tests/etc easier.

blocked by #718


tgummerer (commented):
> pulumi submodule point to master

Note that it doesn't point to `master` (as in it won't auto update), but just pulls in the latest changes from there.  

Also what happens here if the submodule is not checked out?  The build will just not work?  Submodules make me sad :(

brandonpollack23:
> > pulumi submodule point to master
> 
> Note that it doesn't point to `master` (as in it won't auto update), but just pulls in the latest changes from there.
> 
> Also what happens here if the submodule is not checked out? The build will just not work? Submodules make me sad :(

1. Yes, but I have a proposal to bring up in ideation about it
2. #718 fixes that by recursing submodules, but yes locally it is a problem

brandonpollack23:
we discussed this in ideation and decided to go with rennovate bot to update pulumi/pulumi submodule's pointer and the go.mod version to always point to the current master externally, as well as open a PR for it so all the tests (and consequently any new conformance tests) run and fail, letting us know to implement or bypass the test.

PR #718, opened 2025-02-18
recurse submodules on lint.

brandonpollack23:
This is needed so go.mod can point to the /pulumi dir

### DIFF
--- a/.github/workflows/stage-lint.yml
+++ b/.github/workflows/stage-lint.yml
@@ -15,7 +15,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
       - name: Set up Go
         uses: actions/setup-go@v5
         with:


brandonpollack23:
we discussed this in ideation and decided to go with rennovate bot to update pulumi/pulumi submodule's pointer and the go.mod version to always point to the current master externally, as well as open a PR for it so all the tests (and consequently any new conformance tests) run and fail, letting us know to implement or bypass the test.

Frassle:
This isn't needed now? We're planning on not using replace for the go.mod

brandonpollack23:
> This isn't needed now? We're planning on not using replace for the go.mod

not strictly but we should clone the entire repository as we do in the other projects.  This will make it consistent.

brandonpollack23:
> > This isn't needed now? We're planning on not using replace for the go.mod
> 
> not strictly but we should clone the entire repository as we do in the other projects. This will make it consistent.

scratch this, things were failing in other repos without it.  A lot of the actions we do seem to fail lint or some tests without it.

Frassle:
I'm confused? Things are working fine on main without this, what's broken that needs us to do a full clone here, it's just going to slow CI down slightly isn't it?

brandonpollack23:
> I'm confused? Things are working fine on main without this, what's broken that needs us to do a full clone here, it's just going to slow CI down slightly isn't it?

Ill try removing it again maybe it was a previous run or flake

iwahbe:
Things seem to be working.

PR #740, opened 2025-03-12
factor out a `Template` interface and use it in the runner

tgummerer:
In subsequent PRs we're going to introduce components into yaml.  A YAML program will be able to have multiple components, that will be run similar to programs, but won't have all the parts of programs. E.g. a YAML template will be able to include components, but components won't be able to have further components.  Or components won't have packages just for the component itself (but will respect the global packages).

To be able to reuse the runner for running components, factor out a template interface and use it.

No functional changes intended.

Depends on https://github.com/pulumi/pulumi-yaml/pull/739.

### DIFF
--- a/pkg/pulumiyaml/ast/template.go
+++ b/pkg/pulumiyaml/ast/template.go
@@ -482,6 +482,18 @@ func CustomTimeouts(create, update, delete *StringExpr) *CustomTimeoutsDecl {
 	return CustomTimeoutsSyntax(nil, create, update, delete)
 }
 
+type Template interface {
+	GetName() *StringExpr
+	GetDescription() *StringExpr
+	GetConfig() ConfigMapDecl
+	GetVariables() VariablesMapDecl
+	GetResources() ResourcesMapDecl
+	GetOutputs() PropertyMapDecl
+	GetPackages() []packages.PackageDecl
+
+	NewDiagnosticWriter(w io.Writer, width uint, color bool) hcl.DiagnosticWriter
+}
+
 // A TemplateDecl represents a Pulumi YAML template.
 type TemplateDecl struct {
 	source []byte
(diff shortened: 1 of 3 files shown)


pkg/pulumiyaml/ast/template.go
```diff
@@ -534,12 +596,6 @@ func TemplateSyntax(node *syntax.ObjectNode, description *StringExpr, configurat
 	}
 }
 
-func Template(description *StringExpr, configuration ConfigMapDecl, variables VariablesMapDecl, resources ResourcesMapDecl,
```
tgummerer, line 537: This function is unused, and `pkg` is private so no need to worry about backwards compatibility.  Removed it because it clashes with the nice `Template` interface name.

tgummerer:
>  Why can't my component define a sub-component?

I'm not sure that would make a lot of sense. You'd end up with a YAML like

```
name: a-component
runtime: yaml
components:
  aComponent:
    [...]
    components:
      anotherComponent:
        [...]
```

But `aComponent` and `anotherComponent` above would be completely disjoint, and I don't know if it would make sense to have something like that.  I'd rather not allow that right now instead of adding all that additional complexity.

> Why wouldn't components also need to declare packages that are used in the component?

Package declarations would always be global to a yaml plugin, not per component I'd think. Iiuc the packages are defined with the lockfiles in `sdks`, which would always be global.  I imagined any component would be able to use those global packages, but there's no need to define them per component.

The other bit components won't have is the `configuration` stanza (they will only have `config`).  `configuration` is already deprecated in `pulumi-yaml`, and I'd rather not add deprecated things to a new thing.

Frassle:
Ah I follow, this is for the component part of the overall template. Not a separate MLC like component thing.

Frassle (approved):

Issue #745, opened 2025-03-12
Implement schema generation


tgummerer:
Once we can parse components from a yaml file, we need to be able to generate a schema.  That schema can then be used to generate SDKs for various languages for the components.

PR #748, opened 2025-03-13
implement schema generation for yaml components

tgummerer:
For components to work we need to generate their schema.  In YAML we currently only support strings, integers, booleans and arrays as types for config declarations, so let's support the same thing for YAML components for now.

Outputs types are not super straightforward to infer, so right now we're sticking to `pulumi.json/#Any`.  In future revisions we can support more accurate types here.

The schema generation is still not visible to the user in any way, exposing things will be the next step.

Fixes https://github.com/pulumi/pulumi-yaml/issues/745

### DIFF
--- a/pkg/pulumiyaml/ast/template.go
+++ b/pkg/pulumiyaml/ast/template.go
@@ -14,6 +14,7 @@ import (
 	yamldiags "github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/diags"
 	"github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/packages"
 	"github.com/pulumi/pulumi-yaml/pkg/pulumiyaml/syntax"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/pulumiyaml/ast/template_test.go
```diff
@@ -118,3 +118,103 @@ func TestComponentParsing(t *testing.T) {
...
+	syntax, diags := encoding.DecodeYAML("<stdin>", yaml.NewDecoder(strings.NewReader(componentSchemaExample)), nil)
+	require.Len(t, diags, 0)
+
+	template, diags := ParseTemplate([]byte(componentSchemaExample), syntax)
+	require.Len(t, diags, 0)
+
+	spec, err := template.GenerateSchema()
+	require.NoError(t, err)
+
+	require.Equal(t, "yaml-plugin", spec.Name)
```
mikhailshilkov, line 168: Probably a matter of taste, but I'd serialize to JSON and compare JSONs for easier readability

pkg/pulumiyaml/ast/template.go
```diff
@@ -698,6 +699,127 @@ func (d *TemplateDecl) NewDiagnosticWriter(w io.Writer, width uint, color bool)
...
+			"nodejs": schema.RawMessage(`{"respectSchemaVersion": true}`),
+			"python": schema.RawMessage(`{"respectSchemaVersion": true}`),
+			"cshap":  schema.RawMessage(`{"respectSchemaVersion": true}`),
+			"java":   schema.RawMessage(`{"respectSchemaVersion": true}`),
+			"go":     schema.RawMessage(`{"respectSchemaVersion": true}`),
+		},
+	}
+
+	resourcesDef := make(map[string]schema.ResourceSpec)
+	for _, component := range d.Components.Entries {
```
mikhailshilkov, line 748: Did I miss a PR where we introduced d.Components?

pkg/pulumiyaml/ast/template_test.go
```diff
@@ -118,3 +118,103 @@ func TestComponentParsing(t *testing.T) {
...
 }
+
+const componentSchemaExample = `
+name: yaml-plugin
+description: A YAML plugin
+runtime: yaml
+components:
+  aComponent:
+    description: "A component"
+    config:
```
mikhailshilkov, line 129: Do we have to call it `config` not `inputs`?

pkg/pulumiyaml/ast/template.go
```diff
@@ -698,6 +699,127 @@ func (d *TemplateDecl) NewDiagnosticWriter(w io.Writer, width uint, color bool)
...
+			resourceDef.InputProperties[k] = schema.PropertySpec{
+				TypeSpec: typeSpec,
+				Default:  def,
+				DefaultInfo: &schema.DefaultSpec{
+					Environment: []string{k},
+				},
+				Secret: v.Secret != nil && v.Secret.Value,
+			}
+			if def == nil {
+				resourceDef.RequiredInputs = append(resourceDef.RequiredInputs, k)
```
mikhailshilkov, line 780: Are inputs always required if no default is specified? Can a default be a zero value like `""`? I guess it makes sense if so.

pkg/pulumiyaml/ast/template.go
tgummerer: It was introduced here: https://github.com/pulumi/pulumi-yaml/pull/747 Looks like I messed up tagging you there, sorry about that.

pkg/pulumiyaml/ast/template_test.go
tgummerer: It's a very simple change.  I mainly called this `config` to keep the familiarity with current YAML program, but I can change it.  Let's keep that to a separate PR though so we can discuss. (None of this can be used right now, so we're free to change it)

pkg/pulumiyaml/ast/template.go
tgummerer: Yes, a default can be a zero value like that.  As long as `default` is specified it's fine.

mikhailshilkov (approved):
LGTM

Issue #746, opened 2025-03-12
Implement running plugins


tgummerer:
To be able to run a component, we need to be able to run plugins in yaml. Implement `RunPlugin` in the language host and make it able to run components.

Issue #227, opened 2022-05-21
Multiple files for Yaml resources


BoSoeborgPetersen:
## Hello!
- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
I would like to avoid files getting too long, so I would like to split my resources over multiple files, either by using some kind of import statement in the main file, that points to other files like:

**Pulumi.yaml:**
```
name: github-actions-runners-cluster
runtime: yaml
description: GitHub Actions Runners Cluster

resources:
  resourceFile: container.yaml
  resourceGroup:
    type: azure-native:resources:ResourceGroup
```
**Container.yaml**
```
resources:
  registry:
    type: azure-native:containerregistry:Registry
  workspace:
    type: azure-native:operationalinsights:Workspace
  managedCluster:
    type: azure-native:containerservice:ManagedCluster
```

Or implicitly by including everything from all files in a folder called "resources", in addition to the main yaml file.
Or both?

terekete:
Is it possible to do this today ?

nstires-ctgx:
> Is it possible to do this today ?

Doesn't appear so per https://pulumi-community.slack.com/archives/C03E69EQRD2/p1664981326357059

+1 to feature request.

desteves:
+1 

@EronWright

jamest-pin:
I would really like this. My `Pulumi.yaml` file is enormous and unwieldy

AaronFriel:
@terekete @nstires-ctgx @desteves @jamest-pin et al., I regret that I wrote something for this in the community Slack and neglected to update this issue. Using the `compiler` runtime option, a script can be used to merge multiple YAML files. See this repository for an example.

https://github.com/AaronFriel/pulumi-yaml-multi-file

I'm quite partial to this style myself, as I can put the relevant configuration alongside each resource. It's definitely a "Better Helm template", in my opinion. Hopefully this workaround helps manage the complexity of your YAML configurations.

At the same time, please continue to upvote this issue though if an "in-box" solution is important to you!

jamest-pin:
@AaronFriel very cool. At first glance looks like it will do what I need.
Thank you

nikhilsbhat:
This was one of the use cases that led me to build **yamll**.

[...]

PR #749, opened 2025-03-13
implement YAML plugin running

tgummerer:
Allow YAML provider plugins to be run.  Note that this is slightly different than YAML programs in that it natively slurps in all yaml files in the directory, rather than just running the main file.  It merges all `*.yaml` files in the plugin directory instead.

Fixes https://github.com/pulumi/pulumi-yaml/issues/746
Depends on https://github.com/pulumi/pulumi-yaml/pull/748

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -178,19 +178,19 @@ func LoadPluginTemplate(directory string) (*ast.TemplateDecl, syntax.Diagnostics
 				continue
 			}
 			if len(t.Configuration.Entries) > 0 {
-				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `configuration` is not supported in plugins", ""))
+				diags.Extend(syntax.Error(nil, "PulumiPlugin.yaml: root-level `configuration` is not supported in plugins", ""))
 			}
 			if len(t.Config.Entries) > 0 {
-				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `config` is not supported in plugins", ""))
+				diags.Extend(syntax.Error(nil, "PulumiPlugin.yaml: root-level `config` is not supported in plugins", ""))
 			}
 			if len(t.Variables.Entries) > 0 {
-				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `variables` field is not supported in plugins.", ""))
+				diags.Extend(syntax.Error(nil, "PulumiPlugin.yaml: root-level `variables` field is not supported in plugins.", ""))
 			}
 			if len(t.Resources.Entries) > 0 {
-				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `resources` field is not supported in plugins.", ""))
+				diags.Extend(syntax.Error(nil, "PulumiPlugin.yaml: root-level `resources` field is not supported in plugins.", ""))
 			}
 			if len(t.Outputs.Entries) > 0 {
-				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `outputs` field is not supported in plugins.", ""))
+				diags.Extend(syntax.Error(nil, "PulumiPlugin.yaml: root-level `outputs` field is not supported in plugins.", ""))
 			}
 
 			err = template.Merge(t)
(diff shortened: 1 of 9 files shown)


pkg/pulumiyaml/run.go
```diff
@@ -146,6 +146,76 @@ func LoadDir(directory string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
...
+			if t == nil {
+				diags.Extend(syntax.Error(nil, "failed to load template", ""))
+				continue
+			}
+			if len(t.Configuration.Entries) > 0 {
+				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `configuration` is not supported in plugins", ""))
+			}
+			if len(t.Config.Entries) > 0 {
+				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `config` is not supported in plugins", ""))
+			}
```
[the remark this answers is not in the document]
tgummerer: I don't think we support configuring the component provider in other languages currently, and I was leaning on [`ComponentMain`](https://github.com/pulumi/pulumi/blob/master/pkg/resource/provider/component_provider.go#L69) when implementing this, which doesn't support config either, so I cut it out of scope for now.  I'll open an issue as a reminder for us to implement it later.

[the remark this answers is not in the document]
tgummerer: Forgot to mention earlier, I opened https://github.com/pulumi/pulumi-yaml/issues/751 for this.

pkg/pulumiyaml/run.go
```diff
@@ -146,6 +146,76 @@ func LoadDir(directory string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
...
+		}
+		if template == nil {
+			template = t
+		} else {
+			if t == nil {
+				diags.Extend(syntax.Error(nil, "failed to load template", ""))
+				continue
+			}
+			if len(t.Configuration.Entries) > 0 {
+				diags.Extend(syntax.Error(nil, "Pulumi.yaml: root-level `configuration` is not supported in plugins", ""))
```
mikhailshilkov, line 181: PulumiPlugin.yaml?

pkg/pulumiyaml/run.go
```diff
@@ -146,6 +146,76 @@ func LoadDir(directory string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
 	return template, diags, nil
 }
 
+// Load a plugin template from the given directory.
+func LoadPluginTemplate(directory string) (*ast.TemplateDecl, syntax.Diagnostics, error) {
+	// Get all yaml files in the directory, load them and merge them into a single template.
```
mikhailshilkov, line 151: Are we sure we want this now? Are we sure we want to merge all YAML files? How will it work for YAML programs and their stack files, deployment setting files, arbitrary YAML files that users left in their folders, etc?

My preference is probably to keep it to a single file for now, and figure out merging in a separate PR later on the way it would work for any YAML program, not just plugins. I don't feel like we need multi-file for the launch.

mikhailshilkov (commented):
Looks neat! Any testing we could add in addition to this simple integration test?

pkg/pulumiyaml/run.go
Frassle: This is just merging for yaml plugins, that's all new so I'd like to have this. We can work out what we want to do for normal programs later, but it would be good to make them merge as well.

mikhailshilkov: Why do you want to have it? Are we going to recommend users create a file per component? I'd keep it simple and show just PulumiPlugin.yaml in all our docs.

tgummerer: Yeah, I think not merging actually sets us up for more headaches later if we don't do it now.  

> How will it work for YAML programs and their stack files, deployment setting files, arbitrary YAML files that users left in their folders, etc?

Since we're doing this only for plugins and that's new as Fraser mentioned, we don't have to worry about this, because nobody can expect this to work otherwise.  And I think merging yaml files is neat, I remember splitting up YAML files being requested multiple times.

tgummerer:
> Any testing we could add in addition to this simple integration test?

Not sure?  The integration test pretty much covers everything here apart from specific error messages.  Open to suggestions here :)

pkg/pulumiyaml/run.go
tgummerer: Here's the issue requesting this: https://github.com/pulumi/pulumi-yaml/issues/227.  It's the most upvoted issue in this repo with a bunch of comments.  And since we can support it without having to worry about backwards compat now, I think we should just do it.

PR #753, opened 2025-03-18
Fix submodule url scheme

kmosher:
The `git@` scheme will cause clients to try to use ssh, which is unnecessary for a public repo like pulumi/pulumi.

Notably, this will fix Renovate running on this repo.


Frassle (approved):

Frassle:
acceptance test failure isn't due to this PR, we need to regen the conformance test

PR #758, opened 2025-03-19
Support the packages Pulumi.yaml section in YAML runtime

mikhailshilkov:
Update to the latest pulumi/pulumi version that added support for the `packages` section in Pulumi.yaml. We need to thread the packages definition into the YAML loader to make sure it works for the YAML runtime as well.

To test the change, I changed the existing component integration test to replace the `plugins` section with `packages`.

Part of https://github.com/pulumi/pulumi/issues/18918


tgummerer (approved):
Such a simple implementation after the work in pulumi/pulumi. Love it.

mikhailshilkov:
> Such a simple implementation after the work in pulumi/pulumi. Love it.

Well, ideally it would just work with no changes, as for all other runtimes. But YAML is unique in that it has its own package loader, in this sense the required code change is indeed quite limited.

PR #759, opened 2025-03-19
add namespace field to YAML

tgummerer:
Add a namespace field to YAML, which is then generated into the schema, so the user can set it.  This will be used in codegen to distinguish from pulumi packages.

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -73,7 +73,7 @@ func LoadFromCompiler(compiler string, workingDirectory string, env []string) (*
 	}
 
 	templateStr := stdout.String()
-	template, tdiags, err := LoadYAMLBytes(fmt.Sprintf("<stdout from compiler %v>", name), []byte(templateStr), false /* isComponent */)
+	template, tdiags, err := LoadYAMLBytes(fmt.Sprintf("<stdout from compiler %v>", name), []byte(templateStr))
 	diags.Extend(tdiags...)
 	if err != nil {
 		return nil, diags, err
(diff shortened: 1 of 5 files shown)


pkg/pulumiyaml/run.go
```diff
@@ -265,6 +243,9 @@ func LoadYAMLBytes(filename string, source []byte) (*ast.TemplateDecl, syntax.Di
 	if t.Configuration.Entries != nil {
 		diags = append(diags, syntax.Warning(nil, "Pulumi.yaml: root-level `configuration` field is deprecated; please use `config` instead.", ""))
 	}
+	if !isComponent && t.Namespace != nil {
+		diags = append(diags, syntax.Error(nil, "Pulumi.yaml: root-level `namespace` field is only supported in components.", ""))
```
mikhailshilkov, line 247: Wouldn't the right way to frame this as only supported in plugins (PulumiPlugin.yaml not Pulumi.yaml)?

Frassle: Maybe component plugins?

tgummerer: > components could be local in Pulumi.yaml

No they can't be currently. We explicitly disallow that: https://github.com/pulumi/pulumi-yaml/blob/master/pkg/pulumiyaml/run.go#L478.  It's something we could introduce later though, so I'll rename it to `component plugins`.

tgummerer: Hrm linking to that line makes me thing we're probably better off doing that check there, then we don't need to pass around `isComponent` everywhere either.

Issue #737, opened 2025-03-10
Add an integration test for package naming of Git plugins


tgummerer:
Add an integration test for https://github.com/pulumi/pulumi-yaml/pull/734

PR #761, opened 2025-03-20
fix provider type for git based plugins

tgummerer:
https://github.com/pulumi/pulumi-yaml/pull/734 made it so that packages are loaded correctly from disk, but at the same time it also used that name internally, making the provider type look strange (see also https://github.com/pulumi/pulumi/issues/18877).

This fixes that (with the help of the pulumi/pulumi changes in https://github.com/pulumi/pulumi/pull/18951), and also adds a regression test, both for the issue fixed in
https://github.com/pulumi/pulumi-yaml/pull/734, and for the new issue found.  Note that https://github.com/pulumi/pulumi-yaml/pull/734 is reverted here, as it is no longer needed with the new way we're loading plugins, taking the download url into account correctly.

Depends on https://github.com/pulumi/pulumi/pull/18951
Fixes https://github.com/pulumi/pulumi/issues/18877
Fixes https://github.com/pulumi/pulumi-yaml/issues/737


mikhailshilkov (approved):
LGTM

PR #768, opened 2025-03-26
load tagged plugins correctly

tgummerer:
This pulls in the changes from
https://github.com/pulumi/pulumi/pull/19023, to make sure we use the LocalName from the pluginspec to load tagged plugins as well, not only plugins that are based on commit hashes.


lunaris (approved):
LGTM, though this highlights the issues rebuilding plugin names from the filesystem for e.g. `pulumi plugin ls` :thinking:

Issue #764, opened 2025-03-21
Can't set a default value for a component input of type integer


mikhailshilkov:
### What happened?

I want to set a default value for an integer input in a YAML component. I have

```yaml
      rsaBits:
        type: integer
        default: 2048
```

failing with

```
yaml-components:index:SelfSignedCertificate resource 'mycert' has a problem: PulumiPlugin.yaml:18,7-14: type mismatch: default value of type number but type integer was specified;
```

changing to `type: number` fails with `unknown type: number`.

### Example

```
runtime: yaml
name: yaml-components
components:
  SelfSignedCertificate:
    inputs:
      algorithm:
        type: string
        default: RSA
      ecdsaCurve:
        type: string
        default: P224
      dnsName:
        type: string
      validityPeriodHours:
        type: integer
      localValidityPeriodHours:
        type: integer
      rsaBits:
        type: integer
        default: 2048
    resources:
      ca:
        type: tls:index/privateKey:PrivateKey
        properties:
          algorithm: ${algorithm}
          ecdsaCurve: ${ecdsaCurve}
          rsaBits: 2048
      caCert:
        type: tls:index/selfSignedCert:SelfSignedCert
        properties:
          privateKeyPem: ${ca.privateKeyPem}
          isCaCertificate: true
          validityPeriodHours: ${validityPeriodHours}
          subject:
            organization: Example Org
          allowedUses:
            - key_encipherment
            - digital_signature
        options:
          parent: ${ca}
      privateKey:
        type: tls:index/privateKey:PrivateKey
        properties:
          algorithm: ${algorithm}
          ecdsaCurve: ${ecdsaCurve}
          rsaBits: 2048
        options:
          parent: ${ca}
      certRequest:
        type: tls:index/certRequest:CertRequest
        properties:
          privateKeyPem: ${privateKey.privateKeyPem}
          dnsNames:
            - ${dnsName}
          subject:
            organization: Example Org
            commonName: ${dnsName}
        options:
          parent: ${privateKey}
      cert:
        type: tls:index/locallySignedCert:LocallySignedCert
        properties:
          certRequestPem: ${certRequest.certRequestPem}
          caPrivateKeyPem: ${ca.privateKeyPem}
          caCertPem: ${caCert.certPem}
          validityPeriodHours: ${localValidityPeriodHours}
          allowedUses:
            - key_encipherment
            - digital_signature
        options:
          parent: ${certRequest}
    outputs:
      pem: ${cert.certPem}
[...]

PR #771, opened 2025-04-07
allow config defaults to be integers

tgummerer:
In YAML it is possible to set a default value for a config entry. However YAML converts all number looking things into `float64`s/`number`s, which makes the config system reject an integer default value for an integer type.

Since there isn't a great way to distinguish floats and ints in YAML, let the config system continue to parse all numbers into floats. However when applying the default value, we check if it can fit an int, and convert it to an int in that case.

Fixes https://github.com/pulumi/pulumi-yaml/issues/764

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -1211,6 +1211,11 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
 					c.Type.Value, ctypes.ConfigTypes)
 			}
 
+			if t == ctypes.Int && expectedType == ctypes.Number {
+				expectedType = ctypes.Int
+				defaultValue = int(defaultValue.(float64))
+			}
+
 			// We have both a default value and a explicit type. Make sure they
 			// agree.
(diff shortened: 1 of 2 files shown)


pkg/pulumiyaml/run.go
```diff
@@ -1211,6 +1211,11 @@ func (e *programEvaluator) registerConfig(intm configNode) (interface{}, bool) {
 					c.Type.Value, ctypes.ConfigTypes)
 			}
 
+			if t == ctypes.Int && expectedType == ctypes.Number {
```
Frassle, line 1214: Should we not check that the default value is safely castable to int. Like if I say type: int, default: 4.3 I think I should still get an error.

tgummerer: Yeah, you're right. Fixed.

tgummerer:
> I do wonder if we should flip the error check to be on the default value line rather than the type line. At the moment its written that the default value sets the expected type then we error if the type line doesn't match it. Feels more natural that the type line sets the expected type then the default line would error if it didn't match.

Yeah that makes sense to me.  I swapped it now.

Issue #770, opened 2025-04-01
Allow object as resource property


dmfigol-aws:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
I use a cross-language component resource that accepts arguments like this:
```yaml
name: "pulumi-yaml"
cidrs:
  ipv4:
    - cidr: 10.20.0.0/16
    - cidr: 100.64.0.0/26
  ipv6:
    - {}
    - {size: 56}
tags:
  vpcTagKey: vpcTagValue
commonTags:
  commonTagKey: commonTagValue
extraOptions:
  enableDnsHostnames: false
  enableDnsSupport: false
```
I want to be able to use these values in the stack configuration:
```yaml
# Pulumi.dev.yaml
config:
  aws:region: eu-central-1
  aws-native:region: eu-central-1
  vpcArgs:
    name: "pulumi-yaml"
    cidrs:
      ipv4:
        - cidr: 10.20.0.0/16
        - cidr: 100.64.0.0/26
      ipv6:
        - {}
        - {size: 56}
    tags:
      vpcTagKey: vpcTagValue
    commonTags:
      commonTagKey: commonTagValue
    extraOptions:
      enableDnsHostnames: false
      enableDnsSupport: false
```
Unfortunately, the following reference in the project configuration doesn't work:
```yaml
# Pulumi.yaml
name: pulumi-aws-vpc-stacks
description: Pulumi VPC stacks deployed with YAML
runtime: yaml
plugins:
  providers:
    - name: aws-networking
      path: ../pulumi-aws-vpc
resources:
  vpc:
    type: aws-networking:index:VPC
    properties: ${vpcArgs}
outputs:
  vpcId: ${vpc.vpcId}
```
I receive the following error
```
-> % pulumi up
Previewing update (dev)
View in Browser (Ctrl+O): https://app.pulumi.com/<my-org>/pulumi-aws-vpc-stacks/dev/previews/1ad934b1-058f-406c-96aa-21e3dbcc0682
error: failed to discover package requirements: Pulumi.yaml:13,17-27: properties must be an object;
```
This enhancement request is to make such invocation possible.

### Workaround
Reference each key separately, but this is not very scalable
```yaml
# Pulumi.yaml
resources:
  vpc:
    type: aws-networking:index:VPC
    properties:
        name: ${vpcArgs.name}
[...]

PR #779, opened 2025-04-14
Allow resource properties to refer to an object variable 

Frassle:
Fixes https://github.com/pulumi/pulumi-yaml/issues/770

This allows setting the properties of a resource with an object expression rather a fixed object:
```
resources:
  randomPassword:
    type: random:RandomPassword
    properties: ${inputs}
```
Where for example inputs is a config, object variable, or possible the output of another resource.

This is done by allowing either a `PropertyMapDecl` or an `Expr` for the "properties" key. For the `PropertyMap` case nothing changes. When we have an `Expr` we type check that the expression type matches the object shape of the resource. To actually set the properties depends on if the expression resolves as a plain map, in which case we simply copy each value off the map. Or if it resolves as an output, in which case we generate an Apply for each expected input property (given by the resource schema) to copy _that_ field from the map we expect the output to eventually resolve to.

### DIFF
--- a/pkg/pulumiyaml/run_test.go
+++ b/pkg/pulumiyaml/run_test.go
@@ -2659,3 +2659,49 @@ outputs:
 	}, pulumi.WithMocks("projectFoo", "stackDev", mocks))
 	assert.NoError(t, err)
 }
+
+// TestResourceSecretObjectProperties tests we can use a secret object symbol for all the objects properties.
+func TestResourceSecretObjectProperties(t *testing.T) {
+	t.Parallel()
+
+	const text = `
+name: test-yaml
+runtime: yaml
+config:
+  props: {}
+variables:
+  inputs:
+    fn::secret: ${props}
+resources:
+  my-resource:
+    type: test:resource:type
+    properties: ${inputs}
+outputs:
+  result:
+    fn::secret: ${my-resource}
+`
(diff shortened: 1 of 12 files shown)


pkg/tests/transpiled_examples/resource-config-pp/resource-config.pp
```diff
@@ -0,0 +1,24 @@
+config inputs {
```
tgummerer, line 1: Is this used anywhere?

pkg/pulumiyaml/run_test.go
```diff
@@ -2617,3 +2617,91 @@ func TestConflictingEnvVarsMultipleDuplicates(t *testing.T) {
...
+variables:
+  inputs:
+    fn::secret: ${props}
+resources:
+  my-resource:
+    type: test:resource:type
+    properties: ${inputs}
+outputs:
+  result:
+    fn::secret: ${my-resource}
```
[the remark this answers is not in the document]
Frassle: Yeh copy paste from another test, I don't think we need it

pkg/tests/transpiled_examples/resource-config-pp/resource-config.pp
tgummerer: I meant in general, this whole file?  But I see now that it's used for the `example` added above?

PR #787, opened 2025-05-06
Bump Go to 1.24 and bump for security updates

julienp:
Fixes https://github.com/pulumi/pulumi-yaml/security/dependabot/32
Fixes https://github.com/pulumi/pulumi-yaml/security/dependabot/33
Fixes https://github.com/pulumi/pulumi-yaml/security/dependabot/30


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/pulumi/pulumi-yaml
 
-go 1.22
+go 1.24
```
julienp, line 3: AFAIK we never use `pulumi-yaml` as a library, and we're free to bump this.

We need at least 1.23 for the security fixes.

lunaris: [Docsgen](https://github.com/pulumi/registry/blob/master/tools/resourcedocsgen/go.mod) depends on all languages after we hoisted out of `pulumi/pulumi` to break circular dependencies. That said, their toolchain is 1.24 there, so fine?

lunaris (approved):
Comments on docsgen but looks good! Thanks!

PR #797, opened 2025-05-16
Implement `DocLanguageHelper.GetModuleName`

iwahbe:
Implement the interface introduced in https://github.com/pulumi/pulumi/pull/19569.


justinvp (approved):

pkg/pulumiyaml/codegen/doc.go
```diff
@@ -25,6 +25,13 @@ func (d DocLanguageHelper) GetMethodResultName(schema.PackageReference, string,
 	return ""
 }
 
+func (d DocLanguageHelper) GetModuleName(_ schema.PackageReference, modName string) string {
```
justinvp, line 28: Should we add some tests?

PR #798, opened 2025-05-16
Downgrade to go 1.23

t0yv0:
Unfortunately pulumi-terraform-bridge library build-depends on pulumi-yaml as a library and therefore go 1.24 language requirement is a bit more infectious than anticipated and needs to affect all downstream bridged providers. Since pulumi/pulumi is itself on go 1.23 for the moment, this downgrades to 1.23 to match; the idea is to upgrade everything in Pulumi to 1.24 when the time comes to do that.

The HELPMAKEGO utility was automatically upgrading the project to go1.24 this is now fixed with `go install -mod readonly`.

https://github.com/pulumi/pulumi-terraform-bridge/issues/3077 reminder to chase down and remove the build-dependency.

### DIFF
--- a/scripts/install_helpmakego.sh
+++ b/scripts/install_helpmakego.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+HELPMAKEGO_VERSION="$1"
+HELPMAKEGO=bin/${HELPMAKEGO_VERSION}/helpmakego
+
+if ! [ -x ${HELPMAKEGO} ]; then
+    echo "Installing helpmakego@${HELPMAKEGO_VERSION}"
+    go install -mod readonly "github.com/iwahbe/helpmakego@${HELPMAKEGO_VERSION}"
+    mkdir -p "bin/${HELPMAKEGO_VERSION}"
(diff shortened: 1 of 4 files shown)


scripts/install_helpmakego.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+HELPMAKEGO_VERSION="$1"
+HELPMAKEGO=bin/${HELPMAKEGO_VERSION}/helpmakego
+
+if ! [ -x ${HELPMAKEGO} ]; then
+    echo "Installing helpmakego@${HELPMAKEGO_VERSION}"
+    go install -mod readonly "github.com/iwahbe/helpmakego@${HELPMAKEGO_VERSION}"
```
julienp, line 10: I wonder if it works if we set `GOPATH=bin/${HELPMAKEGO_VERSION}/` here to avoid updating the global `$GOPATH/bin/helpmakego` ?

julienp (approved):

scripts/install_helpmakego.sh
t0yv0: Yes found a way to not clobber $GOPATH/bin/helpmakego.

PR #801, opened 2025-05-20
Fix LoadTemplate resolving the project directory to . on Windows

Zaid-Ajaj:
In https://github.com/pulumi/pulumi/pull/19582 we changed the plugin context of the convert process to be the output directory where the generated project lives. The change was simple but suddenly a smoke test started to fail on _Windows_ and it wasn't a flake:
```
=== FAIL: smoke TestYamlConvertSmoke (3.19s)
    environment.go:88: Created new test environment:  C:\Users\RUNNER~1\AppData\Local\Temp\test-env3425171743
    environment.go:197: Running command D:\a\pulumi\pulumi\bin\pulumi plugin install resource random 4.13.0
    environment.go:197: Running command D:\a\pulumi\pulumi\bin\pulumi convert --strict --language pcl --from yaml --out out
    smoke_test.go:[96](https://github.com/pulumi/pulumi/actions/runs/15116997853/job/42498541044?pr=19582#step:41:97): Run Error: exit status 0xffffffff
    smoke_test.go:96: STDOUT: 
    smoke_test.go:96: STDERR: Converting from yaml...
        Downloading provider: yaml
        error: rpc error: code = Unknown desc = load yaml program: reading template Main: open Pulumi.yaml: The system cannot find the file specified.
    smoke_test.go:96: Ran command pulumi args [convert --strict --language pcl --from yaml --out out] and expected success. Instead got failure.
```
Even though the `ConvertProgramRequest` hasn't changed at all and pulumi-converter-yaml does use the `SourceDirectory` correctly. 

After hours of debugging on my personal Windows machine and figuring out how to build everything, I was able to track it down to `LoadTemplate(dir string)` resolving the project directory using `projectDir := path.Dir(projectPath)` where `path.Dir(...)` returned `.` (implicit working directory) instead of using the the dir which is the in the input of the function. 

This problem only now surfaced because before https://github.com/pulumi/pulumi/pull/19582, the implicit working directory was the YAML project directory but now it is the output directory (where PCL is generated)

This PR fixes the problem by not using `path.Dir`, instead using `filepath.Dir`


pkg/pulumiyaml/codegen/eject.go
```diff
@@ -99,9 +98,9 @@ func LoadTemplate(dir string) (*workspace.Project, *ast.TemplateDecl, hcl.Diagno
 
 	main := proj.Main
 	if main == "" {
-		main = projectDir
+		main = dir
```
tgummerer, line 101: I'm confused.  This is a change in behaviour, as it will now look for the `main` in a different directory than before.  What if `projectDir != dir`, it would be incorrect now, no?  

We should probably have some tests for this as well.

Frassle: Yeh this looks wrong. I suspect _dir_ is the problematic part here, something is calling LoadTemplate with the wrong argument.

Frassle:
> using path.Dir(projectPath) is unnecessary because we calculated projectPath from the project directory dir so that dir == path.Dir(projectPath) but without the Windows problems

This isn't true for cases where dir is a subfolder inside the project tree. In that case projectDir will be a parent of dir.

pkg/pulumiyaml/codegen/eject.go
Zaid-Ajaj: I did not know/expect that LoadTemplate would walk up the directory tree to find the project, why does it do that? Will fix the above to not change the behavior

>   I suspect dir is the problematic part here, something is calling LoadTemplate with the wrong argument.

For the issue I am working on, dir is correct (receiving `SourceDirectory` from `ConvertProgramRequest`) it is only on Windows that if `dir == projectDir` then projectDir is `.`

tgummerer: Couldn't you just use `projectDir = filepath.Abs(projectDir)` before the call to `path.Dir`?  That should also solve the windows issue by just making the path an absolute path?

tgummerer: Oh that's dumb, I guess we should never use `path` anywhere then :thinking: Good to know!

pkg/pulumiyaml/codegen/eject_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_loadProjectFromDir(t *testing.T) {
+	t.Parallel()
+	testPath, err := filepath.Abs("./testdata/simple_project")
```
tgummerer, line 14: Maybe use `filepath.Join` here to assemble the path?  Same in the other tests below.

pkg/pulumiyaml/codegen/testdata/nested_project/Pulumi.yaml
```diff
@@ -0,0 +1,2 @@
+name: nested
+runtime: yaml
```
tgummerer, line 2: Missing newlines at the end of the file.  Same for the other files.

tgummerer (approved):

Issue #814, opened 2025-06-12
Panic with Cloudflare program


VenelinMartinov:
### What happened?

Got a panic trying to reproduce a cloudflare provider issue:

```
   provider_program_test.go:82: failed to preview update: failed to run preview: exit status 255
        code: 255
        stdout: Previewing update (test):
          pulumi:pulumi:Stack: (same)
            [urn=urn:pulumi:test::example-ruleset::pulumi:pulumi:Stack::example-ruleset-test]
        panic: interface conversion: ast.Expr is *ast.SymbolExpr, not *ast.StringExpr
        
        goroutine 63 [running]:
        
        github.com/pulumi/pulumi-yaml/pkg/pulumiyaml.(*typeCache).typeExpr(0x1400003b350, 0x1400078c440, {0x105e26180?, 0x14000893a40})
        
        	/home/runner/work/pulumi-yaml/pulumi-yaml/pkg/pulumiyaml/analyser.go:1050 +0xf74
        
        github.com/pulumi/pulumi-yaml/pkg/pulumiyaml.walker.walk({0x1400045ab80, 0x1400045ab70, 0x1400045aba0, 0x1400045ab50, 0x1400045ab90, 0x1400045ab60}, 0x1400078c440, {0x105e26180, 0x14000893a40})
        
        	/home/runner/work/pulumi-yaml/pulumi-yaml/pkg/pulumiyaml/analyser.go:1278 +0x2c0
        
        github.com/pulumi/pulumi-yaml/pkg/pulumiyaml.walker.walk({0x1400045ab80, 0x1400045ab70, 0x1400045aba0, 0x1400045ab50, 0x1400045ab90, 0x1400045ab60}, 0x1400078c440, {0x105e26298, 0x14000893a70})
        
        	/home/runner/work/pulumi-yaml/pulumi-yaml/pkg/pulumiyaml/analyser.go:1254 +0x36c
        
        github.com/pulumi/pulumi-yaml/pkg/pulumiyaml.walker.walk({0x1400045ab80, 0x1400045ab70, 0x1400045aba0, 0x1400045ab50, 0x1400045ab90, 0x1400045ab60}, 0x1400078c440, {0x105e26180, 0x14000893aa0})
        
        	/home/runner/work/pulumi-yaml/pulumi-yaml/pkg/pulumiyaml/analyser.go:1263 +0x424
        
[...]

PR #816, opened 2025-06-12
fix panic when object key is not a string

tgummerer:
We currently don't support object keys that are not strings. However when a user uses something different than a string, e.g. a symbol, we just panic.  We can do better and give the user a diagnostic instead. Do that here.

Fixes https://github.com/pulumi/pulumi-yaml/issues/814

### DIFF
--- a/pkg/pulumiyaml/analyser_test.go
+++ b/pkg/pulumiyaml/analyser_test.go
@@ -367,3 +367,20 @@ func TestConfigCompatibility(t *testing.T) {
 		})
 	}
 }
+
+func TestNonStringKeyInObjectReturnsError(t *testing.T) {
+	tc := typeCache{
+		exprs: make(map[ast.Expr]schema.Type),
+	}
+	expr := &ast.ObjectExpr{
+		Entries: []ast.ObjectProperty{
+			{
+				Key: &ast.BooleanExpr{},
+			},
+		},
+	}
+	_ = tc.typeExpr(nil, expr)
+	require.Equal(t, 1, len(tc.exprs))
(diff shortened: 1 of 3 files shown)


pkg/pulumiyaml/analyser_test.go
```diff
@@ -367,3 +367,22 @@ func TestConfigCompatibility(t *testing.T) {
...
+	expr := &ast.ObjectExpr{
+		Entries: []ast.ObjectProperty{
+			{
+				Key: &ast.BooleanExpr{},
+			},
+		},
+	}
+	_ = tc.typeExpr(nil, expr)
+	require.Equal(t, 1, len(tc.exprs))
+	require.Equal(t, "Object key must be a string, got *ast.BooleanExpr",
```
Frassle, line 386: That's not a friendly error for users. Can we get the user-friendly type of the expression instead? So it would just say "got boolean" here.

tgummerer: Not easily I think.  We do the same thing in other parts of the file: https://github.com/pulumi/pulumi-yaml/pull/816/files#diff-081a538c30d4b80890272ca7cc1aaa9b8e84fc6592814ab06452220440cc0eacR1095-R1108.  I guess we could add an extra method to the Expr interface like `Type()` for this.  Not sure it's worth it though.

Frassle: merge this because it fixes a panic, but we probably should make these errors nicer. I'll raise an issue about it.

Frassle (approved):

PR #832, opened 2025-07-16
[WIP] Include YAML `config` in the template configuration

i-am-tom:


pkg/pulumiyaml/ast/template.go
```diff
@@ -694,7 +699,15 @@ func (d *TemplateDecl) GetConfig() ConfigMapDecl {
 	if d == nil {
 		return ConfigMapDecl{}
 	}
-	// TODO: merge config and configuration (?)
+
```
i-am-tom, line 702: We change `GetConfig` to return `Config` rather than `Configuration`, and then have to update all the references

pkg/pulumiyaml/sort.go
```diff
@@ -120,12 +120,59 @@ func topologicallySortedResources(t ast.Template, externalConfig []configNode) (
...
+		}
 	}
-	for _, node := range append(templateConfig, externalConfig...) {
+
+	for _, node := range externalConfig {
 		cname := node.key().Value
+
+		// Same as the `Config` and `Configuration` cases, except we have to
+		// account for overrides.
+		if _, found := intermediates[cname]; found {
```
i-am-tom, line 162: We have at least one test case where a variable is declared externally _and_ in `config`. Previously, we didn't look at `config`, so this passed just fine. Now, to maintain this behaviour, we allow for external values to override "template" values.

i-am-tom:
Closed in favour of #861.

Issue #840, opened 2025-07-22
Dynamically bridged provider in YAML does not allow explicit providers


VenelinMartinov:
### What happened?

Using a dynamically bridged provider in YAML does not allow an explicit provider to be configured:

```
Diagnostics:
  pulumi:providers:random (random-provider):
    error: Could not automatically download and install resource plugin 'pulumi-resource-random' at version v0.12.0, install the plugin using `pulumi plugin install resource random v0.12.0`: error downloading provider random to file: failed to download plugin: random-0.12.0: 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-random-v0.12.0-darwin-arm64.tar.gz
```

It seems to try to download a normal provider. I also tried it with the TF `null` provider which has no corresponding pulumi provider and got the same error.

### Example

```yaml
name: dynamic_random
description: A minimal Pulumi YAML program
runtime: yaml
packages:
  random:
    source: terraform-provider
    version: 0.12.0
    parameters:
      - hashicorp/random
resources:
  random-provider:
    type: pulumi:providers:random
    defaultProvider: true
  pet:
    type: random:Pet
    properties:
      length: 2
```

1. `pulumi package add terraform-provider hashicorp/random`
2. `pulumi preview`

### Output of `pulumi about`

```
pulumi about
CLI          
Version      3.183.0
Go Version   go1.24.5
Go Compiler  gc

Plugins
KIND      NAME                VERSION
resource  terraform-provider  0.12.0
language  yaml                1.21.2

Host     
OS       darwin
Version  15.5
Arch     arm64

This project is written in yaml

Current Stack: pulumi/dynamic_random/dev

TYPE                     URN
pulumi:pulumi:Stack      urn:pulumi:dev::dynamic_random::pulumi:pulumi:Stack::dynamic_random-dev
pulumi:providers:random  urn:pulumi:dev::dynamic_random::pulumi:providers:random::default_3_6_3
random:index/pet:Pet     urn:pulumi:dev::dynamic_random::random:index/pet:Pet::pet


Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/venelin-pulumi-corp
User           venelin-pulumi-corp
Organizations  venelin-pulumi-corp, demo, pulumi
Token type     personal

Dependencies:
NAME    VERSION
random  3.6.3

[...]

PR #841, opened 2025-07-22
Fix and enable l2-explicit-parameterized-provider

Frassle:
Fixes https://github.com/pulumi/pulumi-yaml/issues/840

When registering resources we didn't look up package refs for provider types correctly, trying to look up a ref for "pulumi" rather than their package because their type tokens are of the form
"pulumi:providers:pkg".

### DIFF
--- a/.changes/unreleased/bug-fixes-842.yaml
+++ b/.changes/unreleased/bug-fixes-842.yaml
@@ -1,0 +1,6 @@
+component: runtime
+kind: bug-fixes
+body: Always set a version for component schemas for `package add`
+time: 2025-07-23T10:52:03.349537+01:00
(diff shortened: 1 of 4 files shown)


cmd/pulumi-language-yaml/language_test.go
```diff
@@ -239,7 +238,6 @@ func TestLanguage(t *testing.T) {
 
 	//nolint:paralleltest // YAML runtime is stateful and not safe to run in parallel.
 	for _, tt := range tests.Tests {
-		tt := tt
```
Frassle, line 242: This isn't need anymore, modern versions of Go already copy the loop variable

PR #848, opened 2025-07-25
Add `object` as a type in `pulumi-yaml`

i-am-tom:
Currently, we run into issues with the latest release because we have no way, as users, to specify that some config variable is an opaque `object`. This causes problems with, for example, the pattern of creating a `props` variable to pass props to other parts of the structure. This PR introduces the `object` type - it's just a `map[string]any` - and allows users to specify it as a type in config(uration). With this, we'll be able to close out #832.


pkg/pulumiyaml/analyser.go
```diff
@@ -361,10 +361,23 @@ func (tc *typeCache) isAssignable(fromExpr ast.Expr, to schema.Type) *notAssigna
 		return okIfAssignable(hasValidEnumValue(fromExpr, to.Elements))
 	case *schema.ObjectType:
 		// We implement structural typing for objects.
-		from, ok := from.(*schema.ObjectType)
-		if !ok {
+		switch f := from.(type) {
```
i-am-tom, line 364: Previously, we implicitly converted objects into maps, but we had no facility to interpret maps as objects. Because our `object` type matches objects with no structural information about its values, we now have to handle this case.

Frassle (approved):

PR #854, opened 2025-08-20
[WIP] Fix incorrect config type

i-am-tom:
I think this is maybe enough of an edge case that it hasn't been discovered, but we incorrectly state that the type of `bool[]` is actually a list of numbers. This PR fixes that.


.changes/unreleased/bug-fixes-854.yaml
```diff
@@ -0,0 +1,6 @@
+component: runtime
+kind: bug-fixes
+body: Lists of booleans now typecheck correctly.
```
tgummerer, line 3:
```suggestion
body: Typecheck lists of booleans correctly
```

tgummerer (approved):
Looks obviously correct

PR #861, opened 2025-09-04
Include `config` in type-checking

### LINKED ISSUE #827: Error: resource, variable, or config value "myValue" not found on v1.21
### What happened?

Going from v1.21.0 to v1.21.1 cause this regression. Attached is a file to help repro it. 
[reproduce-yaml-bug.txt](https://github.com/user-attachments/files/20985089/reproduce-yaml-bug.txt)

### Example

See attached file. 

### Output of `pulumi about`

v3.181.0

### Additional context

_No response_

### Contributing

i-am-tom:
In #827, a bug was reported because we started returning an error that we previously hadn't wired up. This is good because it meant that we caught more bad programs, but bad because it exposed a few bugs in the code that previously weren't visible.

The bug in question happened because we only ever included `configuration` and never `config` in our analysis. The knock-on result is that any variables declared within `config` would be considered undefined elsewhere. This PR simply updates the `GetConfig` method (which previously returned `configuration`) to return `config` and anything in `configuration` whose key wasn't in `config`. In other words, we now return `config` and `configuration` but give priority to `config`.

Another result of this change is that now we have to declare types for variables in `config`: no longer will `props: {}` be accepted because it contains neither a type nor a default value. I've updated the unhappy tests to include both, and all other tests seem to work happily with config maps as before.

I'm a little unsure about this change, so I'm hoping someone can help to validate it. I've spent a _long_ time going round in circles with this trying to understand whether what I'm looking at is a bigger bug (i.e. there's something fundamentally wrong with the way we handle config vs configuration, there's something wrong with the code path for `pulumi convert`), but I think ultimately this fixes the problem with a diff about a thousand lines smaller. If anyone here is a `pulumi-yaml` expert, feedback is very much appreciated.

Fixes #827.


pkg/pulumiyaml/run_test.go
```diff
@@ -2656,7 +2652,10 @@ func TestResourceSecretObjectProperties(t *testing.T) {
 name: test-yaml
 runtime: yaml
 config:
-  props: {}
+  props:
+    type: object
+    default:
+      foo: "bar"
```
tgummerer, line 2658: What happens if you have default `foo: "xyz"` here and pass the config map below?  Does the default get overwritten correctly?

tgummerer (approved):
This does look like the right thing to do to me.  I'm not sure why we didn't merge `config` and `configuration` in the first place here.  

There *might* be some complaints that `config` actually type checks now, but I think that's okay, as we're fixing a bug here.

pkg/tests/integration_test.go
```diff
@@ -63,15 +63,6 @@ func TestInvalidResourceObject(t *testing.T) {
 	})
 }
 
-//nolint:paralleltest // uses parallel programtest
```
i-am-tom, line 66: Behaviour change: previous type mismatches will now be solved by prioritising `config` over `configuration`.

tgummerer: I think this is fine.  `configuration` is deprecated, and having both is just a user error.  I *could* be confusing, but I think that's okay.

pkg/pulumiyaml/run_test.go
```diff
@@ -2618,6 +2618,8 @@ runtime: yaml
 config:
   props:
     type: object
+    default:
+      foo: "bar"
```
tgummerer, line 2622: Does this test need to change?

PR #864, opened 2025-09-08
Fix object type regression

i-am-tom:
When #848 introduced the object type, the analyser code did not correctly decode object values. This PR updates the analyser code to fix this.

Fixes https://github.com/pulumi/pulumi/issues/20468.

### DIFF
--- a/pkg/pulumiyaml/analyser_test.go
+++ b/pkg/pulumiyaml/analyser_test.go
@@ -284,7 +284,6 @@ func TestTypePropertyAccess(t *testing.T) {
 			expectedType: "Invalid",
 			errMsg: `Cannot access into start of type Union<List<string>, Map<number>, {foo: List<any>}>:'start' could be a type that does not support accessing:
   Array<string>: cannot access a property on 'start' (type List<string>)
-  Map<number>: cannot access a property on 'start' (type Map<number>)
   Cannot index via string into 'start.foo' (type List<any>)`,
 		},
 	}


pkg/tests/integration_test.go
```diff
@@ -350,3 +350,20 @@ func TestResourcePropertiesConfig(t *testing.T) {
 		Dir: filepath.Join("testdata", "resource-properties-config"),
 	})
 }
+
+//nolint:paralleltest // uses parallel programtest
+func TestPropertyAccessOnObjects(t *testing.T) {
```
i-am-tom, line 355: Precisely the example given in https://github.com/pulumi/pulumi/issues/20468

.changes/unreleased/bug-fixes-864.yaml
```diff
@@ -0,0 +1,6 @@
+component: runtime
+kind: bug-fixes
+body: Object values are now correctly decoded within YAML programs
```
tgummerer, line 3:
```suggestion
body: Decode object values correctly within YAML programs
```

pkg/pulumiyaml/analyser_test.go
```diff
@@ -284,7 +284,6 @@ func TestTypePropertyAccess(t *testing.T) {
 			expectedType: "Invalid",
 			errMsg: `Cannot access into start of type Union<List<string>, Map<number>, {foo: List<any>}>:'start' could be a type that does not support accessing:
   Array<string>: cannot access a property on 'start' (type List<string>)
-  Map<number>: cannot access a property on 'start' (type Map<number>)
```
i-am-tom, line 287: This error says we can't index into a map. However, we now can, so the error we would encounter (that "start" is not a number) will now come later.

PR #885, opened 2025-11-03
Support scalar invokes

Frassle:
YAML support for https://github.com/pulumi/pulumi/issues/7435

This checks package schemas to see if functions are declared with `ReturnType` rather than `Outputs`, and if so to pick off the single value from the returned property map.

We also error (maybe should just be a warning) that you can't use `return` with an invoke returning a single value.

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -2408,16 +2408,6 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
 		}
 		return output, true
 	})
-				}
-			}
-		}
-
-		output := pulumi.OutputWithDependencies(e.pulumiCtx.Context(), pulumi.Any(retv), dependsOn...)
-		if secret {
-			return pulumi.ToSecret(output), true
-		}
-		return output, true
-	})
 	return performInvoke(args)
 }
 


pkg/pulumiyaml/run.go
```diff
@@ -2359,18 +2360,40 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
...
+		if singleReturnType != nil {
+			if len(result) != 1 {
+				e.addWarnDiag(
+					t.Syntax().Syntax().Range(),
+					fmt.Sprintf("fn::invoke of %s expected a single return value, got: %+v", t.Token.Value, result), "")
+			}
+
+			if t.Return.GetValue() != "" {
+				e.error(t.Return, fmt.Sprintf("fn::invoke of %s has a single return value; cannot specify property '%s'", t.Token.Value, t.Return.Value))
+				return e.error(t.Return, fmt.Sprintf("fn::invoke of %s has a single return value; cannot specify property '%s'", t.Token.Value, t.Return.Value))
```
iwahbe: Duplicate?
```suggestion
				return e.error(t.Return, fmt.Sprintf("fn::invoke of %s has a single return value; cannot specify property '%s'", t.Token.Value, t.Return.Value))
```

pkg/pulumiyaml/run.go
```diff
@@ -2359,18 +2360,40 @@ func (e *programEvaluator) evaluateBuiltinInvoke(t *ast.InvokeExpr) (interface{}
...
+			}
+		} else {
+			if t.Return.GetValue() == "" {
+				retv = result
+			} else {
+				var ok bool
+				retv, ok = result[t.Return.Value]
+				if !ok {
+					e.error(t.Return, fmt.Sprintf("Unable to evaluate result[%v], result is: %+v", t.Return.Value, t.Return))
+					return e.error(t.Return, fmt.Sprintf("fn::invoke of %s did not contain a property '%s' in the returned value", t.Token.Value, t.Return.Value))
```
iwahbe: It feels like one of these should be details for the other:


```suggestion
					e.addDiag(ast.ExprError(expr, 
						fmt.Sprintf("Unable to evaluate result[%v], result is: %+v", t.Return.Value, t.Return),
						fmt.Sprintf("fn::invoke of %s did not contain a property '%s' in the returned value", t.Token.Value, t.Return.Value),
					))
					return nil, false
```

PR #899, opened 2025-12-03
Use the engine to load plugins

### LINKED ISSUE #301: Defer schema loading to engine and remove redundant plugin loads during execution
This is a placeholder issue describing an issue while design(s) are discussed to address.

The Pulumi YAML runtime host loads plugin schemas in order to emulate codegenerated SDK behavior for property defaults & constants, and to type check resources and functions in user programs.

This however results in a discrepancy in behavior with features such as Plugin Link, where the `pulumi` process uses a different method of plugin resolution than the `pulumi-language-yaml` process.
- https://github.com/pulumi/pulumi/pull/10146

We could continue to re-implement or export some of this behavior in a way that Pulumi YAML could reuse, but it places constraints on engine development and leads to a curious situation where the process tree looks like so:

```
pulumi
├─ pulumi-language-yaml
│  ├─ pulumi-resource-foo
│  ├─ pulumi-resource-bar
├─ pulumi-resource-foo
├─ pulumi-resource-bar
```

Every provider plugin is hosted twice, once by the engine process `pulumi`, and also by Pulumi YAML. As some providers can have large memory requirements and because engine process & YAML resolution may differ, it can impact customers and future feature development in the engine.

iwahbe:
We already had this capability in the engine, and we use it for language tests (conformance tests). We should use the same capability during normal operation.

Fixes #301

### DIFF
--- a/.changes/unreleased/Improvements-301.yaml
+++ b/.changes/unreleased/Improvements-301.yaml
@@ -1,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: Use the engine for plugin loading, instead of loading another copy of a plugin for type checking
+time: 2025-12-03T12:49:52.832569+01:00
+custom:
+    PR: "301"
--- a/cmd/pulumi-language-yaml/language_test.go
+++ b/cmd/pulumi-language-yaml/language_test.go
@@ -165,7 +165,7 @@ func TestLanguage(t *testing.T) {
 	// Run the language plugin
 	handle, err := rpcutil.ServeWithOptions(rpcutil.ServeOptions{
 		Init: func(srv *grpc.Server) error {
-			pulumirpc.RegisterLanguageRuntimeServer(srv, server.NewLanguageHost(engineAddress, "", true /* useRPCLoader */))
+			pulumirpc.RegisterLanguageRuntimeServer(srv, server.NewLanguageHost(engineAddress, ""))
 			pulumirpc.RegisterConverterServer(srv, plugin.NewConverterServer(converter.New()))
 			return nil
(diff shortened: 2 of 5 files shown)


.changes/unreleased/Improvements-301.yaml
```diff
@@ -0,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: Use the engine for plugin loading, instead of loading another copy of a plugin for type checking.
```
tgummerer:
```suggestion
body: Use the engine for plugin loading, instead of loading another copy of a plugin for type checking
```

tgummerer (approved):
>  +19 −60 

:drop_of_blood:

iwahbe:
It turns out, the perf cost is from the engine handing the schema over the mapper, not from the gRPC passage itself. [Addressing](https://github.com/pulumi/pulumi/pull/23551) this [slowdown](https://github.com/pulumi/pulumi/issues/23550) allows us to drop plugin loading in YAML without any perf cost.

  | Metric | Before: pulumi-yaml main (in-process loader) | After: PR #899 (engine loader, fixed CLI) |
  |---|---|---|
  | Preview wall time | 2.84s ± 0.07 | 2.87s ± 0.14 |
  | Preview user CPU | 2.70s | 2.74s |
  | Engine `GetSchema` service time | n/a (no RPC) | 1.5ms |
  | Engine peak memory footprint | 46MB | 94MB |

iwahbe:
https://github.com/pulumi/pulumi/pull/23551 has merged and so it's safe to merge this one.

PR #913, opened 2026-01-08
Update pulumi submodule to v3.215.0

julienp:
Use a tag to track this, so that renovate can update this for us in the future.


cmd/pulumi-language-yaml/language_test.go
```diff
@@ -195,6 +195,8 @@ var expectedFailures = map[string]string{
 
 	"l2-resource-option-hide-diffs":          "Currently failing on update to 3.205.0",
 	"l2-resource-option-replacement-trigger": "not yet implemented",
+
+	"provider-resource-component": "requires a provider implemented in the language, not applicable for YAML",
```
Frassle: Rather then adding these here just add an if check in the test runner below that if it starts with "provider-" skip it. Same as we do for "l3-"

Issue #925, opened 2026-01-26
Convert should include aliases


iwahbe:
### What happened?

`l2-resource-option-alias` shows that we don't include aliases when we convert from PCL.

### Example

Given the following PCL:

```
resource "parent" "simple:index:Resource" {
    value = true
}

resource "aliasURN" "simple:index:Resource" {
    value = true
    options {
        parent = parent
        aliases = ["urn:pulumi:test::l2-resource-option-alias::simple:index:Resource::aliasURN"]
    }
}

resource "aliasNewName"  "simple:index:Resource" {
    value = true
    options {
        aliases = [{name = "aliasName"}]
    }
}

resource "aliasNoParent" "simple:index:Resource" { 
    value = true
    options { 
        parent = parent
        aliases = [{noParent = true}]
    }
}

resource "aliasParent" "simple:index:Resource" { 
    value = true
    options {
        parent = parent
        aliases = [{parent = aliasURN}]
    }
}
```

We produce:

```
resources:
  parent:
    type: simple:Resource
    properties:
      value: true
  aliasURN:
    type: simple:Resource
    properties:
      value: true
    options:
      parent: ${parent}
  aliasNewName:
    type: simple:Resource
    properties:
      value: true
    options: {}
  aliasNoParent:
    type: simple:Resource
    properties:
      value: true
    options:
      parent: ${parent}
  aliasParent:
    type: simple:Resource
    properties:
      value: true
    options:
      parent: ${parent}
```

### Output of `pulumi about`

This is on master (9a4a2866d4078843811633485685c0ba4fe5082b) at time of reporting.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Issue #156, opened 2022-04-20
Support full alias set


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Right now, `$resource.options.alias` is a list of alias urns. The other SDKs have a much [more extensive alias control](https://github.com/pulumi/pulumi/blob/f6e5aa5b98c7a8c130cd716926347f18e127cbe1/sdk/go/pulumi/alias.go#L23-L47). This issue tracks brining YAML up to full parity.

PR #936, opened 2026-02-02
Add support for complex alias types

iwahbe:
This PR adds full alias support. Users can now write:

```yaml
resources:
  myResource:
    type: test:resource:type
    properties:
      foo: oof
    options:
      aliases:
        - parent: "someString"
```

Fixes https://github.com/pulumi/pulumi-yaml/issues/925
Fixes https://github.com/pulumi/pulumi-yaml/issues/156

### DIFF
--- a/pkg/pulumiyaml/run.go
+++ b/pkg/pulumiyaml/run.go
@@ -1542,25 +1542,32 @@ func (e *programEvaluator) registerResourceWithParent(kvp resourceNode, parent p
 				evalString := func(value ast.Expr, assignTo *pulumi.StringInput) {
 					val, ok := e.evaluateExpr(value)
 					if ok {
-						if urnStr, ok := val.(string); ok {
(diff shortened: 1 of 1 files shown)


pkg/pulumiyaml/run.go
```diff
@@ -1521,11 +1522,114 @@ func (e *programEvaluator) registerResourceWithParent(kvp resourceNode, parent p
...
+							switch value := value.(type) {
+							case string:
+								parsed, err := urn.Parse(value)
+								if err != nil {
+									e.error(entry.Value, err.Error())
+									overallOk = false
+								} else {
+									alias.ParentURN = pulumi.URN(parsed)
+								}
+							case pulumi.StringOutput:
```
Frassle: This is handling StringOutput and poison, should evalURN also handle that?

PR #960, opened 2026-02-25
Add envVarMappings resource option

guineveresaenger:
Enacts changes made in https://github.com/pulumi/pulumi/pull/21572 upon the YAML runtime.


pkg/pulumiyaml/run.go
```diff
@@ -1828,6 +1828,32 @@ func (e *programEvaluator) registerResource(kvp resourceNode) (lateboundResource
...
+						break
+					}
+					mappings[key] = s
+				}
+				if overallOk {
+					opts = append(opts, pulumi.EnvVarMappings(mappings))
+				}
+			}
+		} else {
+			overallOk = false
```
Frassle: I'd flip this if to be !ok same as the inner if so this short block is read first.

Issue #955, opened 2026-02-21
Components do not correctly parent function invocations


iwahbe:
### What happened?

When defining a component in Pulumi YAML, functions are not correctly parented to their parent component. This causes component resource options (like `providers`) to not apply correctly.

### Example

```yaml
# repro/PulumiPlugin.yaml
name: repro
runtime: yaml
components:
  Component:
    variables:
      ami:
        fn::invoke:
          function: aws:ec2:getAmi
          arguments:
            mostRecent: true
            # Canonical's AWS account ID for official Ubuntu AMIs.
            # https://documentation.ubuntu.com/aws/aws-how-to/instances/find-ubuntu-images/
            owners:
              - "099720109477"
            filters:
              - name: name
                values:
                  - "ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-*"
              - name: virtualization-type
                values:
                  - hvm
    outputs:
        imageId: ${ami.imageId}
```

```yaml
# Pulumi.yaml
name: dev-yaml
runtime: yaml
config:
  pulumi:disable-default-providers: ["aws"]
resources:
  aws:
    type: pulumi:providers:aws
    properties:
      region: us-east-2
  repro:
    type: repro:Component
    options:
      providers:
        - ${aws}
packages:
  repro: ./repro
```

To see the error, run `pulumi install && pulumi up` with ambient AWS credentials. You will see this:

```console
$ pulumi up
Previewing update (iwahbe)

View in Browser (Ctrl+O): https://app.pulumi.com/iwahbe/dev-yaml/iwahbe/previews/210d0601-3085-4f85-8e93-678d4c9a96f3

     Type                      Name             Plan       Info
 +   pulumi:pulumi:Stack       dev-yaml-iwahbe  create     3 errors
 +   ├─ pulumi:providers:aws   aws              create     
 +   └─ repro:index:Component  repro            create     

Diagnostics:
  pulumi:pulumi:Stack (dev-yaml-iwahbe):
    error: rpc error: code = Unknown desc = Invoke: Default provider for 'aws' disabled. 'aws:ec2/getAmi:getAmi' must use an explicit provider.
    
      on PulumiPlugin.yaml line 7:
       7:         fn::invoke:
       8:           function: aws:ec2:getAmi
       9:           arguments:
      10:             mostRecent: true
      11:             # Canonical's AWS account ID for official Ubuntu AMIs.
[...]

PR #961, opened 2026-02-25
Changelog for v1.29.1

julienp:
### Bug Fixes

- [runtime] Correctly apply parent to invokes running within a component [#955](https://github.com/pulumi/pulumi-yaml/pull/955)


tgummerer (approved):
Did you mean to update the submodule in this PR?

PR #975, opened 2026-03-06
Automatically handle new conformance tests from renovate PRs

iwahbe:
This PR:
- adds a new claude skill to handle conformance tests upgrades
- invokes claude on that skill when a new renovate PR is opened

The hope is that this can eliminate the toil of failing pu/pu upgrade PRs from this.


Frassle (approved):

Frassle:
Doing the same for dotnet/java?

PR #1046, opened 2026-04-10
Fix l2-resource-option-version and l2-resource-option-plugin-download-url conformance tests

iwahbe:
Four bugs prevented per-resource version/pluginDownloadURL options from working correctly:

1. **Codegen (PCL → YAML)**: `genResourceOpts` did not emit `version` or `pluginDownloadURL` resource options, so generated YAML programs lost these settings.

2. **Eject (YAML → PCL)**: The importer applied the *highest* package version to every resource of that package instead of preserving the per-resource version. Fixed to use each resource's own version option, falling back to the latest package version only when unset.

3. **Runtime**: When a resource specifies an explicit version or pluginDownloadURL, the SDK-registered packageRef must be cleared so the engine builds a provider request from the per-resource options rather than using the SDK's fixed version.

4. **Shared descriptor mutation**: `loadPackage` mutated the shared `packageDescriptors` map entry when overriding the version, which could affect subsequent resources. Fixed by copying the descriptor before mutation.

Also simplifies `GetReferencedPackages` to not emit error diagnostics on version conflicts between resources — per-resource version options are runtime directives, not package-level conflicts.


pkg/pulumiyaml/packages.go
```diff
@@ -267,10 +253,11 @@ func loadPackage(
 			Version:     version,
 			DownloadURL: pluginDownloadURL,
 		}
-	}
-	if version != nil {
-		// Override the version if one was passed in.
-		descriptor.Version = version
+	} else if version != nil {
+		// Override the version if one was passed in, but don't mutate the shared descriptor.
```
Frassle, line 257: Why do we override version but not pluginDownloadURL? That feels odd, like it should be both or neither.

iwahbe: It should be both. Fixed and I've added a test.

Frassle (approved):

PR #1057, opened 2026-04-21
Changelog for v1.32.0

pgavlin:


.changes/v1.32.0.md
```diff
@@ -0,0 +1,10 @@
+## v1.32.0 - 2026-04-21
+
+### Improvements
+
+- [codegen] update golang to 1.25.9 [#1044](https://github.com/pulumi/pulumi-yaml/pull/1044)
+
+### Bug Fixes
+
+- [runtime, codegen, convert] Fix per-resource version and pluginDownloadURL options to correctly create distinct providers [#<no value>](https://github.com/pulumi/pulumi-yaml/pull/<no value>)
```
Frassle: Another missing value... Is this AI making up the changelog entries and skipping the issue number part?

pgavlin: I'll check.

pgavlin: (I ran all of this manually, so I don't see how it could be?)

pgavlin: I think I understand your question better now--are you wondering if the LLM is making up the bits that `changie` consumes from `.changes`?

pgavlin: (fixed, in any case--also added a note to CONTRIBUTING)

Frassle (approved):

PR #1069, opened 2026-04-29
Fix l1-builtin-list

Frassle:
This required adding `singleOrNone` and `length`. Also found that `split` was wrong, we'll want to get a focused conformance test in place for the string functions I think.


iwahbe (approved):

iwahbe:
You need to regenerate `TestGenerateProgram/join-template`.

PR #1087, opened 2026-05-08
Fix l2-resource-option-custom-timeouts

iwahbe:
## Summary

- Widens `CustomTimeoutsDecl` fields (`Create`, `Update`, `Delete`) from
  `*StringExpr` to `Expr` so `customTimeouts` can take interpolations and
  config references (e.g. `${createTimeout}`), not just string literals.
- Runtime evaluates each timeout expression and validates that it
  resolves to a string before passing it to `pulumi.Timeouts`.
- Codegen import (YAML → PCL) now routes timeout values through
  `importExpr`, preserving expressions across a round-trip.
- Removes `l2-resource-option-custom-timeouts` from the conformance
  test expected-failures list.

## Test plan

- [x] `go test -run TestLanguage ./cmd/pulumi-language-yaml/...`
- [x] `make lint`
- [x] `go test ./pkg/pulumiyaml/...`


cmd/pulumi-language-yaml/testdata/eject-pcl/l2-resource-option-custom-timeouts/program.pp
```diff
@@ -1,3 +1,7 @@
+config createTimeout string {
+	__logicalName = "createTimeout"
```
Frassle, line 2: Why are we still writing out __logicalName for config blocks? They're deprecated

cmd/pulumi-language-yaml/testdata/eject-pcl/l2-resource-option-custom-timeouts/program.pp
```diff
@@ -48,3 +52,14 @@
 		}
 	}
 }
+
+resource configTimeout "simple:index:Resource" {
+	__logicalName = "configTimeout"
```
Frassle, line 57: and this is fine I guess, but a bit redundant to write out a logical name the same as the resource name

Frassle (approved):

PR #1099, opened 2026-05-14
Snippet support

Frassle:
This adds `ConvertSnippet` to the yaml converter. We bind the input file and attributes as mini yaml documents and then run them through a subset of the same process as for a normal `ConvertProgram` call.


tgummerer (approved):

i-am-tom (approved):
It looks sensible

PR #1143, opened 2026-06-30
Drop the pulumi/pulumi submodule

iwahbe:
## Summary

Removes the `pulumi/pulumi` submodule, which was only used to source the synthetic test schemas (`other`, `std`, `using-dashes`) for the codegen and transpile tests.

Those schemas are now embedded in `github.com/pulumi/pulumi/pkg/v3/codegen/testing/utils` (see pulumi/pulumi#23738), so the schema-only plugin host resolves them through the Go module. The host still falls back to the on-disk testdata directory for the real-provider schemas (`aws`, `awsx`, `kubernetes`) that `get_schemas.sh` downloads from provider releases.

## Changes

- Drop the submodule (`.gitmodules` + gitlink).
- Bump `pkg/v3` / `sdk/v3` to the embed-test-schemas build of pulumi/pulumi.
- `get_schemas.sh` no longer fetches the synthetic schemas via the submodule.
- Register the codegen test providers explicitly, since `utils.NewContext`'s
  defaults no longer include `aws`/`awsx`/`kubernetes`.
- Drop the now-unused `submodules: recursive` checkout from CI.

### DIFF
--- a/.github/workflows/stage-test.yml
+++ b/.github/workflows/stage-test.yml
@@ -71,7 +71,6 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v4
         with:
-          submodules: recursive
           ref: ${{ inputs.commit-ref }}
       - name: Install Go
         uses: actions/setup-go@v5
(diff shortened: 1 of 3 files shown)


julienp (approved):
nice!

tgummerer (approved):
Kill all the submodules 🩸

PR #1146, opened 2026-07-07
Support extension parameterization packages

guineveresaenger:
This pull request adds YAML support for extension parameterized packages.

 Part of https://github.com/pulumi/pulumi/issues/17059.


iwahbe (approved):
Don't we have a language test for this? Can we enable it for YAML?

PR #1161, opened 2026-07-29
dump pulumi stack on hangs

### LINKED ISSUE #1158: Workflow failure: Periodically capture coverage
## Workflow Failure

[Periodically capture coverage](https://github.com/pulumi/pulumi-yaml/blob/main/.github/workflows/periodic-cover.yml) has failed. See the list of failures below:

- [2026-07-29T00:58:53.000Z](https://github.com/pulumi/pulumi-yaml/actions/runs/30410659794)
  - ❌ [test / Test](https://github.com/pulumi/pulumi-yaml/actions/runs/30410659794/job/90445746174): Test with Coverage (failure)

tgummerer:
YAML tests seem to hang sometimes for unknown reasons.  The stack traces we have are only for the test harness, and not for the pulumi process, so they are unfortunately not very useful in understanding where exactly we hang.  Add some more stack dumps from the actual pulumi process so we can debug things better.

Closes #1158 (doesn't fix it, but I'd like a new issue opened if it happens again after this is merged)

### DIFF
--- a/cmd/pulumi-language-yaml/language_test.go
+++ b/cmd/pulumi-language-yaml/language_test.go
@@ -16,13 +16,17 @@ package main
 
 import (
 	"bufio"
+	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"strings"
 	"sync"
+	"syscall"
 	"testing"
+	"time"
 
 	"github.com/pulumi/pulumi-yaml/pkg/converter"
 	"github.com/pulumi/pulumi-yaml/pkg/server"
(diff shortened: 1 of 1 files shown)


Frassle:
Why do we think this is a yaml bug if it's an engine side hang? Wouldn't we expect to see that in the other languages conformance tests as well? 

At any rate, should we just put this in the conformance test code, just give like every test like 10 minutes to run at most and do the stack trace dump if that times out? That'll help with any other languages that get stuck as well.

tgummerer:
Tbh I don't really know where the bug is.  I haven't seen the timeout anywhere else other than yaml, but I also haven't found anything in the backtrace we've gotten from the test harness to suggest what the issue could be, so I was hoping this would help.  In fact it looks like the test in  the PR might be timing out, so maybe we don't even need to merge it.

tgummerer:
This did its job.

PR #1162, opened 2026-07-30
Fix package resolution bug for provider extensions

guineveresaenger:
While building e2e tests for KUbernetes CRD extensions, it turned up that in YAML, package resolution was dropping the base package on extensions.

This pull request fixes that and adds unit tests. The reason conformance tests weren't catching this is because conformance tests only cover locks that declare the base package explicitly.

### DIFF
--- a/pkg/pulumiyaml/packages.go
+++ b/pkg/pulumiyaml/packages.go
@@ -255,13 +255,24 @@ func loadPackages(
 
 	packageName := ResolvePkgName(typeString)
 	nsDescriptors := descriptors[tokens.Package(packageName)]
-	if len(nsDescriptors) == 0 {
-		// Fall back to just the package name and passed in version if we don't have a descriptor.
-		nsDescriptors = []*schema.PackageDescriptor{{
+	// For parameterized extensions, add the base provider as an additional resolution candidate.
+	needsBase := true
+	for _, d := range nsDescriptors {
+		if d.Parameterization == nil { // plain unparameterized provider
+			needsBase = false
+			break
+		}
+		if d.Parameterization.Name == packageName { // replacement parameterized provider
+			needsBase = false
+			break
+		}
+	}
+	if needsBase {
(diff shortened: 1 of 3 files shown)


pkg/pulumiyaml/packages.go
```diff
@@ -255,13 +255,24 @@ func loadPackages(
...
-		// Fall back to just the package name and passed in version if we don't have a descriptor.
-		nsDescriptors = []*schema.PackageDescriptor{{
+	// For parameterized extensions, add the base provider as an additional resolution candidate.
+	needsBase := true
+	for _, d := range nsDescriptors {
+		if d.Parameterization == nil { // plain unparameterized provider
+			needsBase = false
+			break
+		}
+		if d.Parameterization.Name == packageName { // replacement parameterized provider
```
Frassle, line 265: This can be one if check `d.Parameterization == nil || d.Parameterization.Name == packageName`

Frassle (commented):
Can you explain more fully why a conformance test can't catch this? I'd have thought conformance would be able to check package resolution was correct.

pkg/pulumiyaml/packages.go
guineveresaenger: the reason I left it separate is so the different provider situations are clear - `d.Parameterization.Name == packageName` meaning "this is a replacement-parameterized provider" is a bit opaque. 🤷‍♀️

guineveresaenger:
@Frassle yeah - so I _think_ what happened is that when the _tests_ call [pcl.BindDirectory](https://github.com/pulumi/pulumi/blob/master/pkg/testing/pulumi-test-language/runner/runner.go#L1222), it returns with both the base _and_ the extension packages already bound. But `pulumi package add` only generates the extension package, so there's no base package to reference for YAML specifically. I wasn't sure if it would make sense or would be worth it to add this scenario to conf tests but if we feel it's a good idea I can do that.
When I reference the `kubernetes` package in my YAML program explicitly, this bug doesn't fire either.

Frassle (approved):

PR #1191, opened 2026-09-15
Remove dead program codegen test cases and golden files

iwahbe:
The skip list in `TestGenerateProgram` names 22 test directories that no longer exist in the pulumi/pulumi program test driver, so those cases can never match. The fake `PULUMI_CONFIG` setup and most of the `testMonitor` mocks served only those removed tests. The `synthetic-resource-properties` case pointed at issue 229, but pulumi/pulumi deleted that test in v3.227.0.

Remove the dead cases, mocks, and config, and delete the 58 golden directories under `pkg/pulumiyaml/testing/test/testdata` that no test reads any more. The nine subtests that ran before still run.


tgummerer (approved):
🩸

